PR #3, opened 2023-05-07
Changes from the nitro repo contracts-develop branch

PlasmaPower:
These changes were forgotten about after being merged into the nitro repo's `contracts-develop` branch. The only functional change is adding the `isSequencer` mapping to the `SequencerInbox`, which is used off-chain to check if a sequencer feed signer is valid.


PlasmaPower:
I haven't found a way yet. It's tricky to do that across multiple branches it seems.

gzeoneth:
> I haven't found a way yet. It's tricky to do that across multiple branches it seems.

Do you think there would be licensing issue if we remove credit to external contributors?

gzeoneth (approved):
LGTM

PR #5, opened 2023-05-11
Use create2 to deploy the RollupProxy and predict its address

PlasmaPower:
This helps simplify the L3 deployment process and prevent issues if many rollups are being deployed at the same time from the same RollupCreator.

I had to move the constructor arguments into an initialize function because otherwise they'd be part of the initcode and they aren't predictable.


gzeoneth (changes requested):
Attacker can still front-run the deployment to deploy a rollup with alternative config at the same address, we should include the rollup config in the salt.

src/rollup/RollupCreator.sol
```diff
@@ -60,20 +65,23 @@ contract RollupCreator is Ownable {
...
     // Bridge should have a single inbox and outbox
-    function createRollup(Config memory config, address expectedRollupAddr)
-        external
-        returns (address)
-    {
+    function createRollup(Config memory config) external returns (address) {
         CreateRollupFrame memory frame;
         frame.admin = new ProxyAdmin();
 
+        frame.rollupSalt = bytes32(deployedRollups++);
```
gzeoneth, line 80: should we just let people to pick a salt instead? this seems to same as using a nonce which can still cause the same issue

yahgwai: Yeh, We might also want to hash the config and contract dependencies into the salt as well to ensure it couldnt be frontrun with different params?

PlasmaPower: Hmm, the part I was trying to achieve is that you don't need to know the rollup address ahead of time, and you can instead just query the receipt logs to get that. Is there another use case where a user would want to know the rollup address ahead of time? We could perhaps make the salt a hash of their address and a nonce they specify.

gzeoneth:
pushed some change to this PR so that the rollup address is deterministic based on the supplied config

PlasmaPower:
I think this is good. My two concerns were if someone wanted to deploy multiple rollups with the same config, which they shouldn't because the chain id and owner are part of the config, and that the deployer would need to serialize the rollup config to predict the address which is complicated in Go, but I realized it shouldn't need to predict the address because it can just get it from the receipt logs.

DZGoldman (approved):

gzeoneth (approved):

PR #6, opened 2023-05-11
Store rollup node ArbSys block numbers if on an Arbitrum host chain

PlasmaPower:
This lets the nitro node do efficient log queries for the node creation event, even when running an L3.


gzeoneth (approved):
We can also change createdAtBlock to return L1/L2 block number based on if it is a L2/L3 to optimize storage, createdAtBlock it is currently used in
1. off-chain event lookup
2. validatorAfk check (which should be replaced by deadlineBlock instead)
3. time delta of minimumAssertionPeriod (can be removed in new protocol)

but LGTM otherwise

PlasmaPower:
Yeah, I had the same thought about the validatorAfk check but didn't realize the minimumAssertionPeriod is getting removed in the new challenge protocol. That's good to hear, but until then I think it'd make sense to merge this as-is.

gzeoneth:
> Yeah, I had the same thought about the validatorAfk check but didn't realize the minimumAssertionPeriod is getting removed in the new challenge protocol. That's good to hear, but until then I think it'd make sense to merge this as-is.

because there is only 1 deterministic assertion to be made, so I think we don't need that anymore; not sure about potential dos tho

PR #7, opened 2023-05-12
Add a CI step to check the PR's base branch

PlasmaPower:
This PR adds a CI check that PRs targeting the main branch as their base must be from either the develop branch or a branch name starting with "hotfix".

To test out this PR, I'm initially opening it against main, and then I'll change it to be against develop and see if it changes from failing to passing. -- Edit: this seems to have worked, but it required pushing another commit to refresh the CI. I think that's an acceptable tradeoff and I don't see a way around it.


gzeoneth:
@PlasmaPower can you force merge this to main instead of develop?

gzeoneth (approved):

PR #8, opened 2023-05-15
add serialized chain config to init message data

magicxyyz:
* adds chain config string to init message data
* adds chain config string to rollup config
* adds `setChainConfig` method to ArbOwner interface

### DIFF
--- a/src/rollup/RollupEventInbox.sol
+++ b/src/rollup/RollupEventInbox.sol
@@ -30,8 +30,12 @@ contract RollupEventInbox is IRollupEventInbox, IDelayedMessageProvider, Delegat
         rollup = address(_bridge.rollup());
     }
 
-    function rollupInitialized(uint256 chainId) external override onlyRollup {
-        bytes memory initMsg = abi.encodePacked(chainId);
+    function rollupInitialized(uint256 chainId, string calldata chainConfig)
+        external
+        override
+        onlyRollup
+    {
+        bytes memory initMsg = abi.encodePacked(chainId, uint8(0), chainConfig);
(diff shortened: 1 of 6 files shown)


src/rollup/RollupEventInbox.sol
```diff
@@ -30,8 +30,12 @@ contract RollupEventInbox is IRollupEventInbox, IDelayedMessageProvider, Delegat
...
     }
 
-    function rollupInitialized(uint256 chainId) external override onlyRollup {
-        bytes memory initMsg = abi.encodePacked(chainId);
+    function rollupInitialized(uint256 chainId, string calldata chainConfig)
+        external
+        override
+        onlyRollup
+    {
+        bytes memory initMsg = abi.encodePacked(chainId, uint8(0), chainConfig);
```
PlasmaPower, line 38: Should we `require(chainConfig.length > 0, "EMPTY_CHAIN_CONFIG");` here? I think the nitro node fails to parse the init message otherwise. Of course we can't fully validate it but I think this is a sensible check.

gzeoneth: what is the uint8(0) here btw?

PlasmaPower: Just a version byte -- if we want to change the init message in the future, it'd be good to have this as an easy way to indicate the version

magicxyyz: you're right, it's safer to have the config checked, thanks for catching it

PlasmaPower (dismissed):
LGTM

gzeoneth (dismissed):
LGTM

PR #15, opened 2023-05-26
chore(ISequencerInbox): Add Missing views

shotaronowhere:
The external view function stubs are missing from the ISequencerInbox interface for

- maxTimeVariation
- dasKeySetInfo

In solidity, public struct [autogenerate](https://docs.soliditylang.org/en/v0.8.20/contracts.html#visibility-and-getters) public views returning the struct members in order (excluding arrays and mappings). In the case of dasKeySetInfo which is a mapping, the view takes a key parameter of the same type as the mapping.

This PR adds the missing stubs.


gzeoneth (approved):
LGTM, thanks @shotaronowhere

PR #16, opened 2023-05-28
adding deployment script

GreatSoshiant:


scripts/deployment.ts
```diff
@@ -0,0 +1,156 @@
...
+    /* 
+      * Call the createRollup function
+      * Adding 15 million gas limit otherwise it'll be reverted with "gas exceeds block gas limit" error
+    */
+   console.log("Calling createRollup to generate a new rollup ...")
+    const createRollupTx = await contracts.rollupCreator.createRollup(rollupConfig, {gasLimit: ethers.BigNumber.from("15000000")});
+    const createRollupReceipt = await createRollupTx.wait();
+    const rollupCreatedEvent = createRollupReceipt.events?.find(
+      (event: { event: string }) => event.event === "RollupCreated"
+      );
```
gzeoneth, line 121: should also match the event address == rollupCreator

scripts/deployment.ts
```diff
@@ -0,0 +1,156 @@
...
+    throw new Error("Private key is not defined");
+  }
+
+  const providerAPI = process.env.RPC_URL;
+  if (!providerAPI) {
+    throw new Error("RPC URL is not defined");
+  }
+
+  const provider = new providers.JsonRpcProvider(providerAPI)
+  const signer = new Wallet(privateKey, provider)
```
gzeoneth, line 67: it appear that this script is expected to be called via hardhat, then why don't we use the hardhat network and signer?

gzeoneth (changes requested):
added some comments also please run `yarn format`

scripts/deployment.ts
GreatSoshiant: I think using this safer, in case we add other events?

scripts/deployment.ts
GreatSoshiant: I think we want to use it to deploy on sepolia. Am I wrong?

scripts/deployment.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  try {
+
+    // Deploying all contracts
+    const contracts = await deployAllContracts(signer);
+    
+    /* 
+      * Call setTemplates with the deployed contract addresses
+      * Adding 15 million gas limit otherwise it'll be reverted with "gas exceeds block gas limit" error
```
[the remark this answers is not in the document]
GreatSoshiant: no, it's on Arb Goerli Harry

scripts/deployment.ts
gzeoneth: Not sure what u mean, at current state it might be possible to fake a RollupCreated event from another contract.

scripts/deployment.ts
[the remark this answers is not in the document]
gzeoneth: Would be nice to figure out why, i.e. if that is a hardhat or rpc issue.

scripts/deployment.ts
gzeoneth: It doesn't matter, you can still config network in hardhat

scripts/deployment.ts
GreatSoshiant: added check here Harry:
https://github.com/OffchainLabs/nitro-contracts/pull/16

scripts/deployment.ts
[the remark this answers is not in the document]
GreatSoshiant: Checked it again Harry! Not sure what happened but it's resolved haha

scripts/deployment.ts
```diff
@@ -0,0 +1,179 @@
...
+  const contract: Contract = await connectedFactory.deploy()
+  await contract.deployTransaction.wait()
+  console.log(`New ${contractName} created at address:`, contract.address)
+  return contract
+}
+
+async function deployAllContracts(
+  signer: any
+): Promise<Record<string, Contract>> {
+  const bridgeCreator = await deployContract('BridgeCreator', signer)
```
gzeoneth, line 33: Missing these bridgeCreator initial template addresses
https://github.com/OffchainLabs/nitro-contracts/blob/5eb3ff91b009263004167230895387dcb30965d6/src/rollup/BridgeCreator.sol#L28-L32

gzeoneth (changes requested):

scripts/deployment.ts
```diff
@@ -0,0 +1,179 @@
...
+        futureSeconds: 7200,
+      },
+    }
+
+    /*
+     * Call the createRollup function
+     * Adding 15 million gas limit otherwise it'll be reverted with "gas exceeds block gas limit" error
+     */
+    console.log('Calling createRollup to generate a new rollup ...')
+    const createRollupTx = await contracts.rollupCreator.createRollup(
```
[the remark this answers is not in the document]
GreatSoshiant: I think it's better to have another script just for creating rollups. I'll do it Harry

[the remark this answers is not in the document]
GreatSoshiant: I've changed the flow Harry. Now there's two different scripts one for deployment, the other for rollup creation. 
Also added a new config.ts to put the rollupConfig there. This isn't a good way maybe to have a config.ts and not adding it to .gitignore, Not sure what would be the best way to deal with it though.
I'll post it on Slack to have visibility on this as well

scripts/rollupCreation.ts
```diff
@@ -0,0 +1,124 @@
...
+          console.log(`Contract RollupProxy is already verified.`)
+        } else {
+          console.error(
+            `Verification for RollupProxy failed with the following error: ${error.message}`
+          )
+        }
+      }
+      console.log('Inbox (proxy) Contract created at address:', inboxAddress)
+      console.log(
+        'Outbox Contract created at address:',
```
gzeoneth, line 91:
```suggestion
        'Outbox (proxy) created at address:',
```

scripts/config.ts
```diff
@@ -0,0 +1,21 @@
...
+  chainId: ethers.BigNumber.from(''),
+  chainConfig: '',
+  genesisBlockNum: ethers.BigNumber.from(''),
+  sequencerInboxMaxTimeVariation: {
+    delayBlocks: ethers.BigNumber.from(''),
+    futureBlocks: ethers.BigNumber.from(''),
+    delaySeconds: ethers.BigNumber.from(''),
+    futureSeconds: ethers.BigNumber.from(''),
+  },
+}
```
gzeoneth, line 21: should populate some devnet dummy but valid value

scripts/deployment.ts
```diff
@@ -1,32 +1,62 @@
...
+): Promise<void> {
+  try {
+    // Define the verification options with possible 'contract' property
+    const verificationOptions: {
+      contract?: string
+      address: string
+      constructorArguments: any[]
+    } = {
+      address: contractAddress,
+      constructorArguments: constructorArguments,
```
gzeoneth, line 21: Let's also specify the contract to verify here, I think it is
```suggestion
      contract: contractName,
      address: contractAddress,
      constructorArguments: constructorArguments,
```
but please test

GreatSoshiant: We don't need it for all contracts Harry. The only contract that had to different codes with same bytecode in the repo, was bridge contract.
I put contract as optional here, because otherwise we need to put the path for all other contracts which makes the code unclean and also we don't need it, cause for other contracts, hardhat was able to find the right contract just by name.

gzeoneth: wdym? This is just passing the contract name right? The current setup is a little but unsafe since say you pass in 0xabcd and Inbox, it will verify and output "Inbox contract verified" even if 0xabcd is Outbox because hardhat will just find the matching contract from everything as you said.

gzeoneth: Or you must use the full path here? Idk

GreatSoshiant: Checked it Harry, and it works well, cause it checks bytecodes, and not names.
We can add paths as well, but because we have different contracts with different paths, it'll make the code so ugly. So my vote would be not doing it cause it works well!
I will resolve this cm, if you think we need to add it please open the discussion again
Thanks!

gzeoneth (dismissed):
lgtm, can you add action entries to packages.json too?

GreatSoshiant:
Yeah for sure Harry 🫡

package.json
```diff
@@ -24,7 +24,10 @@
     "build:0.7": "INTERFACE_TESTER_SOLC_VERSION=0.7.0 yarn run build",
     "test:compatibility": "yarn run build:0.6 && yarn run build:0.7",
     "test:storage": "./test/storage/test.bash",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package",
+    "deployment": "hardhat run scripts/deployment.ts",
+    "rollup": "hardhat run scripts/rollupCreation.ts"
```
gzeoneth, line 29:
```suggestion
    "deploy-factory": "hardhat run scripts/deployment.ts",
    "deploy-rollup": "hardhat run scripts/rollupCreation.ts"
```

gzeoneth (approved):

PR #20, opened 2023-06-05
chore(precompiles): ArbGasInfo interface comment update

shotaronowhere:
The comment about returning zero for the second return param is false and misleading.

https://github.com/OffchainLabs/nitro/blob/ce5eb97644e03586175d0b4b3f6ed9966ac42f70/precompiles/ArbGasInfo.go#L146

The precompile returns the perBlockGasLimit which is the same as the maxTxGasLimit.

### DIFF
--- a/src/precompiles/ArbAddressTable.sol
+++ b/src/precompiles/ArbAddressTable.sol
@@ -29,32 +29,32 @@ interface ArbAddressTable {
      * @param offset offset of target address
      * @return resulting address and updated offset into the buffer (revert if buffer is too short)
      */
-    function decompress(bytes calldata buf, uint256 offset)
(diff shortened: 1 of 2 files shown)


shotaronowhere:
Just adding this change as well, the arb address table return types are wrong. In fact, the [precompile](https://github.com/OffchainLabs/nitro/blob/0b32740ddd245ff0e52b1fd1a0372b90195f6a0c/precompiles/ArbAddressTable.go#L48) will throw if passed the indicated uint256 type for an index. 

This PR just corrects the interface type to uint64.

shotaronowhere:
PR is stale, comments are already addressed.

PR #26, opened 2023-06-16
Include L1 gas charges in L3 batch posting reports

PlasmaPower:


src/bridge/SequencerInbox.sol
```diff
@@ -394,6 +399,12 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
                 seqMessageIndex,
                 block.basefee
             );
+            if (hostChainIsArbitrum) {
+                // Include extra gas for the host chain's L1 gas charging
+                uint256 l1Fees = ArbGasInfo(address(0x6c)).getCurrentTxL1GasFees();
+                uint64 extraGas = uint64(l1Fees / block.basefee);
```
gzeoneth, line 405: this cast is not safe, can we use uint256?

PlasmaPower: I've replaced this code with your recommendation of modifying the data gas cost

gzeoneth (approved):
lgtm

PR #27, opened 2023-06-16
Include the current data cost in the init message

PlasmaPower:
Points into https://github.com/OffchainLabs/nitro-contracts/pull/26 because it uses the ArbitrumChecker


gzeoneth (approved):
LGTM

PR #30, opened 2023-06-22
Add additional redeem checks to mock Simple.sol

PlasmaPower:
These checks catch a bug in retryable gas estimation but should always pass.


gzeoneth (approved):
LGTM

PR #34, opened 2023-07-01
updating rollup created event

GreatSoshiant:
Adding outbox, validatorUtils, and validatorWalletCreator fields to the RollupCreated event. This will help to have these contract addresses easier when a new rollup chain is deployed, for instance on Orbit


src/rollup/RollupCreator.sol
```diff
@@ -112,9 +115,12 @@ contract RollupCreator is Ownable {
         emit RollupCreated(
```
gzeoneth, line 115: let's also emit `rollupEventInbox`, `challengeManager`

or maybe we just emit `(rollup, ContractDependencies, proxyAdmin)`

GreatSoshiant: Thanks Harry!
Added both to the event.

gzeoneth (approved):
LGTM

PR #36, opened 2023-07-04
feat: add immutable maxDataSize to seqInbox

gzeoneth:


src/bridge/Inbox.sol
```diff
@@ -610,8 +610,9 @@ contract Inbox is DelegateCallAware, PausableUpgradeable, IInbox {
         address _sender,
         bytes memory _messageData
     ) internal returns (uint256) {
-        if (_messageData.length > MAX_DATA_SIZE)
-            revert DataTooLarge(_messageData.length, MAX_DATA_SIZE);
+        uint256 maxDataSize = sequencerInbox.maxDataSize();
```
yahgwai, line 613: This does a .call right? That will make all messages 2k more expensive, are we ok with that?

gzeoneth: It does, I am also thinking we can have the same immutable in the inbox and validate it with the sequencer inbox on init or if the sequencerInbox is changed (which I am also looking at implementing right now)

PlasmaPower: The system doesn't break if the max data sizes are different, though realistically we'd always want them to be the same I think

gzeoneth:
https://github.com/OffchainLabs/nitro-contracts/pull/58

PR #37, opened 2023-07-06
feat: setting validators and batch poster in rollup creation

GreatSoshiant:
Now we can set validators and batch poster as inputs of createRollup function in the RollupCreator contract


src/rollup/RollupCreator.sol
```diff
@@ -98,6 +103,8 @@ contract RollupCreator is Ownable {
             osp
         );
 
+        address tempOwner = config.owner;
+        config.owner = address(this);
```
gzeoneth, line 107: confusing name, maybe `actualOwner` instead of `tempOwner`? also add some comment e.g.
```suggestion
        // initialize the rollup with this contract as owner to set batch poster and validators
        // it will transfer the ownership back to the actual owner later
        address actualOwner = config.owner;
        config.owner = address(this);
```

src/rollup/RollupCreator.sol
```diff
@@ -36,6 +37,8 @@ contract RollupCreator is Ownable {
     address public validatorUtils;
     address public validatorWalletCreator;
 
+    bool[1] private vals = [true];
+
```
gzeoneth, line 41:
```suggestion
```

src/rollup/RollupCreator.sol
```diff
@@ -113,6 +120,20 @@ contract RollupCreator is Ownable {
                 validatorWalletCreator: validatorWalletCreator
             })
         );
+        {
+            IRollupAdmin rollupAdmin = IRollupAdmin(address(rollup));
```
gzeoneth, line 124: if you remove this var and just use these below
```
IRollupAdmin(address(rollup)).setValidator(...)
and
IRollupAdmin(address(rollup)).setOwner(...)
```
it should fit into the stack without scoping

edit: you also need to fix the `proxyAdmin.transferOwnership(tempOwner);`

gzeoneth (changes requested):

src/rollup/RollupCreator.sol
```diff
@@ -114,6 +121,19 @@ contract RollupCreator is Ownable {
...
+        // Call setValidator on the newly created rollup contract
+        bool[] memory _vals = new bool[](_validators.length);
+        for (uint256 i = 0; i < _validators.length; i++) {
+            _vals[i] = true;
+        }
+        IRollupAdmin(address(rollup)).setValidator(_validators, _vals);
+
+        IRollupAdmin(address(rollup)).setOwner(actualOwner);
+
+        proxyAdmin.transferOwnership(actualOwner);
```
gzeoneth, line 135: you should move it back to the original place before config.owner is changed and use config.owner

GreatSoshiant: Done

gzeoneth: did you forgot to push?

GreatSoshiant: LOL! Yup! WTF sorry!

src/rollup/RollupCreator.sol
```diff
@@ -114,6 +123,17 @@ contract RollupCreator is Ownable {
...
         );
 
+        sequencerInbox.setIsBatchPoster(_batchPoster, true);
+
+        // Call setValidator on the newly created rollup contract
+        bool[] memory _vals = new bool[](_validators.length);
+        for (uint256 i = 0; i < _validators.length; i++) {
+            _vals[i] = true;
+        }
+        IRollupAdmin(address(rollup)).setValidator(_validators, _vals);
```
PlasmaPower, line 133: We should add a check to not call setValidator if the validator list is empty, because it'll revert otherwise. We should also not set the _batchPoster if it's the zero address.

GreatSoshiant: is it enough to just check the lenght of validators @PlasmaPower ? Or need to check if all addresses are non-zero as well?

PR #40, opened 2023-07-13
Use single RollupCreator for both Eth and Erc20 based rollups

gvladika:
RollupCreator implementation was separated for Eth and Erc20 based rollups. This commit merges it back together. Main differences compared to "original" rollup creator are:

- owner sets 2 bridge creators - `BridgeCreator ethBridgeCreator` and  `ERC20BridgeCreator erc20BridgeCreator`
- `createRollup` receives additional param - `address _nativeToken`
- if provided address is address(0) then `ethBridgeCreator ` is used, otherwise `erc20BridgeCreator`
- `address nativeToken` is emitted as part of `RollupCreated`


gzeoneth (approved):
nice, lgtm

PR #44, opened 2023-07-20
Add getL1RewardRate and getL1RewardRecipient methods to ArbGasInfo interface

ganeshvanahalli:


src/precompiles/ArbGasInfo.sol
```diff
@@ -95,6 +95,12 @@ interface ArbGasInfo {
     /// @notice Get how slowly ArbOS updates its estimate of the L1 basefee
     function getL1BaseFeeEstimateInertia() external view returns (uint64);
 
+    /// @notice Get the L1 pricer reward rate
```
PlasmaPower, line 98: We should document what this rate is measured in. I think in terms of wei per data unit (approx equal to extra wei per L1 calldata gas), but I could be wrong.

ganeshvanahalli: Yes, it is measured in "wei per unit"

PlasmaPower (approved):
LGTM

PR #49, opened 2023-08-03
Add upgrade executor to rollup creation flow

gvladika:
Each call to `createRollup` will create UpgradeExecutor. Executor will be owner of proxyAdmin which owns all rollup and bridge contracts. RollupOwner is the single account that has `executor` role on the UpgradeExecutor.


gzeoneth (changes requested):

src/rollup/RollupCreator.sol
```diff
@@ -91,6 +95,20 @@ contract RollupCreator is Ownable {
...
+                new TransparentUpgradeableProxy(
+                    address(upgradeExecutorLogic),
+                    address(proxyAdmin),
+                    bytes("")
+                )
+            )
+        );
+        address[] memory upgradeExecutors = new address[](1);
+        upgradeExecutors[0] = config.owner;
+        upgradeExecutor.initialize(address(upgradeExecutor), upgradeExecutors);
```
gzeoneth, line 110: let's call this executors for less confusion
```suggestion
        address[] memory executors = new address[](1);
        executors[0] = config.owner;
        upgradeExecutor.initialize(address(upgradeExecutor), executors);
```

src/libraries/UpgradeExecutor.sol
gzeoneth, line 1: Wonder if this should live here or the gov repo, I think it make sense to live here to avoid circular dependency of 2 packages tho we need to make sure the correct OZ version is used. This contract is audited with OZ 4.7.2 in gov but `nitro-contracts` repo is using 4.5.0/4.5.2

gzeoneth: I see little risk in this specific instance but we should still align the version just in case

gvladika: A lot of contracts in this repo are using OZ, so if we changed version now it would be big impact. Shall we simply re-audit this version of UpgradeExecutor? Audit will happen soon anyways

gzeoneth:
pushed a few commit to use the upgrade-executor npm package, kinda tricky because typechain does not include a typesafe deploy function so we cannot use the UpgradeExecutor__factory directly; instead I am using the abi and bytecode to create a factory manually https://github.com/OffchainLabs/nitro-contracts/pull/49/commits/ca401d12769fe5a18e87ebe5d0c2f3b403be0c0e

src/libraries/UpgradeExecutor.sol
gvladika: Now using latest release of upgrade-executor npm package

gzeoneth (approved):

PR #50, opened 2023-08-17
Add sha256 preimage support

PlasmaPower:
As a proof-of-concept for adding more preimage types to the one step proof, this adds SHA256 support.


joshuacolvin0 (approved):
LGTM

PR #51, opened 2023-08-22
feat: helper for keyless deployment using delayed inbox

gzeoneth:
This PR included a new helper contract `DeployHelper` that fund and create delayed message from L1 -> L2 to deploy several keyless deployment contracts. These contract usually will fail deployment on Arbitrum Chains due to the L1 gas factor, and in the worst case the nonce will be burnt which make them impossible to be deployed again. 

When using the delayed inbox, L1 gas is not charged on L2 since the user already paid on L1 to put the message into the delayed inbox and therefore those deployment will succeed.

TODO: move this into rollupCreator as a standard part of rollup deployment so we never miss? downside being deploying them require around 0.13 native token (ETH)


src/rollup/DeployHelper.sol
```diff
@@ -0,0 +1,72 @@
...
+    uint256 internal constant MAXFEEPERGAS = 1000000000;
+
+    function _fundAndDeploy(
+        IInbox inbox,
+        uint256 _value,
+        address _l2Address,
+        bytes memory payload
+    ) internal {
+        uint256 submissionCost = inbox.calculateRetryableSubmissionFee(0, block.basefee);
+        inbox.createRetryableTicket{value: _value + submissionCost + GASLIMIT * MAXFEEPERGAS}({
```
gvladika, line 39: Why do you use `createRetryableTicket` with no payload + `sendL2Message`? If it's only to bypass validation could `depositEth` + `sendL2Message` be more efficient?

gvladika: oh I see, you have to fund specific account

scripts/rollupCreation.ts
```diff
@@ -79,6 +90,11 @@ async function main() {
       const validatorWalletCreator =
         rollupCreatedEvent.args?.validatorWalletCreator
 
+      await DeployHelper__factory.connect(deployHelperAddress, signer).perform(
+        inboxAddress,
+        { value: ethers.utils.parseEther('0.1') }
```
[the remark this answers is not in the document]
gzeoneth: https://github.com/OffchainLabs/nitro-contracts/pull/51/commits/e0af7fba9b2c01dbadeef667226bd410c9433d9a

PR #53, opened 2023-09-01
Bump Hardhat version

gvladika:
Change package version: 
- node version 18 in CI
- Hardhat 2.17.2 (latest)

Bump of `solidity-coverage` required as well due to backward incompatiblity. Latest version `solidity-coverage` still failed:
```
ParserError: Expected ',' but got ';'
   --> src/mocks/Simple.sol:205:40:
    |
205 |         (c_83b71b6f(0x137081c2a34587c9); /* statement */
    |                                        ^


Error in plugin solidity-coverage: HardhatError: HH600: Compilation failed
```  

Solution was to remove parenthesis around low level call in `Simple.sol`. And then use `hardhat-ignore-warnings` to silence 'unused return call value' compiler warning.

Also bumped `hardhat-deploy` and `hardhat-gas-reporter`although it was not neccessary.


gzeoneth (commented):
lgtm, I think we can rebase this on main

PR #59, opened 2023-09-06
Deploy create2 factories as part of rollup creation

gvladika:
This commit adds deployment of L2 CREATE2 factories to the standard process of deploying new rollup. L2 factories are deployed by using DeployHelper contract (originally added by @gzeoneth  here https://github.com/OffchainLabs/nitro-contracts/pull/51) , using retryable tickets. 

One implication is that user needs to pay for retryables, either in ETH or in native fee token. If paying with ETH user simply sends value when calling `createRollup` and excess is refunded back to caller. In case of fee token, user needs to pre-approve RollupCreator to spend the expected fee amount. There is also an option to opt out of deploying L2 factories.


src/rollup/DeployHelper.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    uint256 internal constant GASLIMIT = 100_000;
+    uint256 internal constant MAXFEEPERGAS = 1_000_000_000;
```
yahgwai, line 45: I have this comment https://github.com/OffchainLabs/nitro-contracts/pull/51#discussion_r1307443800 on another pr about maybe supplying this as params

src/rollup/RollupCreator.sol
```diff
@@ -89,22 +100,36 @@ contract RollupCreator is Ownable {
...
+     * @dev - Rollup should be the owner of bridge
+     * @dev - RollupOwner should be the owner of Rollup's ProxyAdmin
+     * @dev - RollupOwner should be the owner of Rollup
+     * @dev - Bridge should have a single inbox and outbox
+     * @dev - Validators and batch poster should be set if provided
+     * @param config       The configuration for the rollup
+     * @param _batchPoster The address of the batch poster, not used when set to zero address
+     * @param _validators  The list of validator addresses, not used when set to empty list
+     * @param _nativeToken Address of the custom fee token used by rollup. If rollup is ETH-based address(0) should be provided
+     * @param _deployFactoriesToL2 Whether to deploy L2 factories using retryable tickets. If true, retryables need to be paid for in native currency.
```
yahgwai, line 119: I think we should include more comments here about the consequences of not deploying the factories

gvladika: https://github.com/OffchainLabs/nitro-contracts/pull/59/commits/116d6515349f01736b67ba3870ac6535d08150fd

src/rollup/DeployHelper.sol
gvladika: implemented here https://github.com/OffchainLabs/nitro-contracts/pull/59/commits/03c98c03c954366dfeab61a21068e6764ad3588a 

maxFeePerGas is provided as param to DeployHelper,  and also to RollupCreator when new rollup is to be created

scripts/rollupCreation.ts
```diff
@@ -86,6 +113,8 @@ export async function createRollup(feeToken?: string) {
 
       console.log("Congratulations! 🎉🎉🎉 All DONE! Here's your addresses:")
       console.log('RollupProxy Contract created at address:', rollupAddress)
+      console.log('Wait a minute before starting the contract verification')
+      await sleep(2 * 60 * 1000)
```
gzeoneth, line 117: waited 2 minutes

src/rollup/DeployHelper.sol
```diff
@@ -0,0 +1,149 @@
...
+
+    uint256 internal constant GASLIMIT = 100_000;
```
gzeoneth, line 44: in retrospect I think this would work with 21000

gzeoneth (approved):

scripts/rollupCreation.ts
gvladika: indeed, fixed

src/rollup/DeployHelper.sol
gvladika: right, we're only doing ETH transfer with 1st retryable. Updated

src/rollup/RollupCreator.sol
yahgwai: Isnt there also the risk that someone else, other than the original deployer, could send these transactions and burn the nonce without

src/rollup/RollupCreator.sol
```diff
@@ -82,30 +93,24 @@ contract RollupCreator is Ownable {
      * @param _batchPoster The address of the batch poster, not used when set to zero address
      * @param _validators  The list of validator addresses, not used when set to empty list
      * @param _nativeToken Address of the custom fee token used by rollup. If rollup is ETH-based address(0) should be provided
+     * @param _deployFactoriesToL2 Whether to deploy L2 factories using retryable tickets. If true, retryables need to be paid for in native currency.
+     *                             Deploying factories via retyrable tickets at rollup creation time is the most reliable method to do it since it
```
yahgwai, line 97:
```suggestion
     *                             Deploying factories via retryable tickets at rollup creation time is the most reliable method to do it since it
```

src/rollup/RollupCreator.sol
gvladika: yep, I mean it doesn't matter who triggers the transaction, problem is if it fails and nonce gets burned. I updated comment to be more clear about that.

PR #65, opened 2023-09-26
Normalize deposited native token to match 18 decimals

gvladika:
There is a scenario where custom fee token has number of decimals different than 18. That's not ideal because a lot of tooling assumes that native currency (on child chain) is using 18 decimals. For that reason we do a normalization, and de-normalization, when depositing and withdrawing native asset through native bridge.

Ie. let's say deposit amount is 752. 
 - if token has 16 decimals normalized amount which is going to be minted on child chain is 75200 
 - if token has 20 decimals normalized amount is 7
 - if token uses no decimals normalized amount is 752*10^18

When withdrawing, opposite process is done. There can be loss of precision, due to the fact that we have to round down amount to the nearest supported value.

We also add 2 requirements on the custom fee token in order to avoid overflow:
- fee token can use at most 36 decimals (enforced at inbox/outbox initialization)
- max amount that can be bridged is `type(uint256).max / 10**18` which is ~1.1*10^59 tokens


yahgwai:
> * if token uses no decimals normalized amount is 752*10^18

Shouldnt this be "if token uses no decimals normalized amount is 752"? Since the decimals are the same.

src/libraries/BytesLib.sol
```diff
@@ -0,0 +1,59 @@
...
+
+/*
+ * @title Solidity Bytes Arrays Utils
+ * @author Gonçalo Sá <goncalo.sa@consensys.net>
+ *
+ * @dev Bytes tightly packed arrays utility library for ethereum contracts written in Solidity.
+ *      The library lets you concatenate, slice and type cast bytes arrays both in memory and storage.
+ */
+
+pragma solidity ^0.8.0;
```
yahgwai, line 11: Add additional comment and link to the token-bridge-contracts file

src/libraries/DecimalsNormalizationHelper.sol
```diff
@@ -0,0 +1,72 @@
...
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+library DecimalsNormalizationHelper {
+    /// @notice normalize the amount as if token had 18 decimals.
+    /// @dev Ie. let's say amount is 752. If token has 16 decimals normalized
+    ///      amount is 75200. If token has 20 decimals normalized amount is 7.
+    ///      If token uses no decimals normalized amount is 752*10^18.
+    /// @param amount amount to normalize
+    /// @return amount normalized to 18 decimals
+    function normalizeAmountTo18Decimals(address token, uint256 amount)
```
yahgwai, line 17: Could maybe share some code by refactoring a bit:

```
    function adjustDecimals(uint256 amount, uint8 decimalsIn, uint8 decimalsOut) internal pure returns (uint256) {
        if(decimalsIn < decimalsOut) {
            return amount * 10**(decimalsOut - decimalsIn);
        } else if(decimalsOut < decimalsIn) {
            return amount / 10**(decimalsIn - decimalsOut);
        } else {
            return amount;
        }
    }

    function normalizeAmountTo18Decimals2(address token, uint256 amount)
        internal
        view
        returns (uint256)
    { 
        // get decimals
        uint8 tokenDecimals = getDecimals(token);

        return adjustDecimals(amount, tokenDecimals, 18);
    }

    function convertToNativeTokenDecimals2(address token, uint256 amount)
        internal
        view
        returns (uint256)
    {
        // get decimals
        uint8 tokenDecimals = getDecimals(token);
        
        return adjustDecimals(amount, 18, tokenDecimals);
    }
```

src/libraries/DecimalsNormalizationHelper.sol
```diff
@@ -0,0 +1,72 @@
...
+
+        return amount;
+    }
+
+    /// @notice convert the amount from normalized 18 decimals back to token's actual number of decimals.
+    /// @dev This process is opposite to the one used in normalization function. It's important to notice
+    ///      that amount is always rounded down when conversion is performed.
+    /// @param amount amount to convert
+    /// @return amount normalized to 18 decimals
+    function convertToNativeTokenDecimals(address token, uint256 amount)
```
yahgwai, line 40: Could maybe rename these, I keep mixing up convert and normalise. It might be simpler just to call them `fromNativeTo18Decimals` and `from18ToNativeDecimals`. And adjust the comments accordingly.

gvladika:
> > * if token uses no decimals normalized amount is 752*10^18
> 
> Shouldnt this be "if token uses no decimals normalized amount is 752"? Since the decimals are the same.

We decided to handle failure to get decimals as 0 decimals. This is in sync with how fullstack team handles it in frontend

src/bridge/ERC20Inbox.sol
```diff
@@ -44,11 +45,22 @@ contract ERC20Inbox is AbsInbox, IERC20Inbox {
             dest = AddressAliasHelper.applyL1ToL2Alias(msg.sender);
         }
 
+        // In order to keep compatibility of child chain's native currency with external 3rd party tooling we
```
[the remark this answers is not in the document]
gvladika: Yes, implemented here - https://github.com/OffchainLabs/nitro-contracts/pull/65/commits/752e16dee6787dccfe7e8e88c2b49ec6fc6df3d2

src/libraries/BytesLib.sol
gvladika: https://github.com/OffchainLabs/nitro-contracts/pull/65/commits/257476acbb097bc73ecb70855bdefee5d82199ee

src/libraries/DecimalsNormalizationHelper.sol
gvladika: https://github.com/OffchainLabs/nitro-contracts/pull/65/commits/b3731e53489f34adaf7103dd46bac15237e15af6

src/libraries/DecimalsNormalizationHelper.sol
gvladika: makes sense, done

src/libraries/BytesLib.sol
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * This file is copied over from offchainlabs/token-bridge-contracts repo.
+ * Ref: https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/libraries/BytesLib.sol
```
yahgwai, line 15: We should probably reference a specific commit rather than main

src/libraries/DecimalsConverterHelper.sol
```diff
@@ -0,0 +1,49 @@
...
+    /// @dev Ie. let's say amount is 752. If token has 16 decimals and is being adjusted to
+    ///      18 decimals then amount will be 75200. If token has 20 decimals adjusted amount
+    ///      is 7. If token uses no decimals converted amount is 752*10^18.
+    ///      When amount is adjusted from 18 decimals back to native token decimals, opposite
+    ///      process is performed.
+    /// @param amount amount to convert
+    /// @param decimalsIn current decimals
+    /// @param decimalsOut target decimals
+    /// @return amount converted to 'decimalsOut' decimals
+    function adjustDecimals(
```
yahgwai, line 21: We should probably add some short tests for the specific functions in this file. It's ok if you want to do that in another PR though

src/libraries/DecimalsConverterHelper.sol
```diff
@@ -0,0 +1,49 @@
...
+        uint256 amount,
+        uint8 decimalsIn,
+        uint8 decimalsOut
+    ) internal pure returns (uint256) {
+        if (decimalsIn < decimalsOut) {
```
yahgwai, line 26: We could make this very slightly more efficient for most cases by adding a decimalsIn == decimalsOut as the first if statement. It wont save much gas though, so find to leave as is as well.

gvladika: done

src/libraries/DecimalsConverterHelper.sol
gvladika: added separate test file with unit tests

yahgwai (approved):

src/bridge/ERC20Inbox.sol
```diff
@@ -139,4 +164,21 @@ contract ERC20Inbox is AbsInbox, IERC20Inbox {
...
+        // expect 18 decimals to be always used for native currency. If native token uses different number of
+        // decimals then here it will be normalized to 18. Keep in mind, when withdrawing from child chain back
+        // to parent chain then the amount has to match native token's granularity, otherwise it will be rounded
+        // down.
+
+        // Also make sure that inflated amount does not overflow uint256
+        if (nativeTokenDecimals < 18) {
+            if (value > MAX_BRIDGEABLE_AMOUNT) {
+                revert AmountTooLarge(value);
```
gzeoneth, line 179: would be nice to also enforce this in nitro so that withdrawal that exceed this size (unlikely) would not be stuck

src/bridge/ERC20Outbox.sol
```diff
@@ -5,11 +5,41 @@
...
 
 import "./AbsOutbox.sol";
+import {IERC20Bridge} from "./IERC20Bridge.sol";
+import {DecimalsConverterHelper} from "../libraries/DecimalsConverterHelper.sol";
+import {AmountTooLarge, NativeTokenDecimalsTooLarge} from "../libraries/Error.sol";
 
 contract ERC20Outbox is AbsOutbox {
-    // it is assumed that arb-os never assigns this value to a valid leaf to be redeemed
+    /// @dev number of decimals used by native token
+    uint8 public nativeTokenDecimals;
```
gzeoneth, line 14: Looks like this can be stored in [ERC20Bridge packed with nativeToken](https://github.com/OffchainLabs/nitro-contracts/blob/686c997cb82b95a6b7d4008613d750572d320c91/src/bridge/ERC20Bridge.sol#L23) to save gas (1 cold sload vs 1 warm call) ; also does not need a initializer

src/bridge/ERC20Outbox.sol
```diff
@@ -22,6 +52,18 @@ contract ERC20Outbox is AbsOutbox {
         return AMOUNT_DEFAULT_CONTEXT;
     }
 
+    /// @inheritdoc AbsOutbox
+    function _getAmountToUnlock(uint256 value) internal view override returns (uint256) {
+        // make sure that inflated amount does not overflow uint256
+        if (nativeTokenDecimals > 18) {
+            if (value > MAX_BRIDGEABLE_AMOUNT) {
```
gzeoneth, line 59: should swap these since MAX_BRIDGEABLE_AMOUNT is a constant (cheaper to access)

also can cache nativeTokenDecimals since it as accessed again in L64

gzeoneth (commented):
LGTM but not a fan of this approach due to additional complexity and increased gas usage. It can be quite confusing when erc20 decimals does not match native decimals (e.g. a withdrawAndCall will need to make sure they converted the decimal if that is also in calldata), and rounding error will be stuck in the bridge.

PR #67, opened 2023-10-01
Fix the BridgeCreator.sol contract

tsite:
Clean up the code and reduce deployment gas cost by moving contract creation outside of the contructor.

Move the sequencer inbox maxTimeVariation from memory to calldata.

Index the sender for MessageDelivered events - this allows bridges to efficiently query the deposit history for an address without requiring an external indexer.

Add DISABLE_VERIFICATION environment variable to disable contract verification.


src/bridge/IBridge.sol
```diff
@@ -13,7 +13,7 @@ interface IBridge {
         bytes32 indexed beforeInboxAcc,
         address inbox,
         uint8 kind,
-        address sender,
+        address indexed sender,
```
gzeoneth, line 16: would prefer to have this in another PR for consideration separately due to how it changes the existing abi

src/rollup/BridgeCreator.sol
```diff
@@ -20,177 +20,106 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 
 contract BridgeCreator is Ownable {
-    ContractTemplates public ethBasedTemplates;
-    ContractERC20Templates public erc20BasedTemplates;
+    BridgeContracts public ethBasedTemplates;
+    BridgeContracts public erc20BasedTemplates;
```
gvladika, line 24: Nice refactor of BridgeCreator, can you also reflect the changes to Foundry unit tests `BridgeCreatorTest` and `RollupCreatorTest`? They are part of the CI loop

tsite: updated the unit tests

gzeoneth:
@cla-bot check

gvladika:
@tsite I cherry-picked your commits, fixed some conflicts and PR has been merged here https://github.com/OffchainLabs/nitro-contracts/pull/77. We'll have latest version of code released soon

PR #68, opened 2023-10-04
Optimisations in the sequencer inbox

yahgwai:
* Don't use proxy ~44k
* Immutable max time variation, as well as <s>rollup and</s> bridge address ~10k
* postUpgradeInit required, see https://github.com/OffchainLabs/nitro-contracts/pull/87

Nitro changes required:
* SequencerBatchDelivered now emitted from the bridge, also the event args have changed to include the sequencer inbox address.
* The sequencer inbox is now immutable, so we will now deploy a new one when changes are required in the sequencer inbox
* The variant of addSequencerL2BatchFromOrigin without the `prevMessageCount`, `newMessageCount` and parameters have been removed, however it will continue to exist on the old sequencer inbox which wont be updated to have it removed. If this function abi was used for decoding historical logs we should instead do this with standard abi decoding.


src/bridge/SequencerInbox.sol
```diff
@@ -181,40 +196,6 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
         );
     }
 
-    /// @dev Deprecated in favor of the variant specifying message counts for consistency
-    function addSequencerL2BatchFromOrigin(
```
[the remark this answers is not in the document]
gzeoneth: we can also have an empty reverting function

gzeoneth (approved):

src/bridge/SequencerInbox.sol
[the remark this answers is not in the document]
yahgwai: Probably easier - I'll change to reverting empty

src/bridge/SequencerInbox.sol
```diff
@@ -73,26 +75,27 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+        uint256 _maxDataSize
+    ) {
+        if (bridge_ == IBridge(address(0))) revert HadZeroInit();
+        bridge = bridge_;
+        rollup = bridge_.rollup();
+        if (address(rollup) == address(0)) revert RollupNotChanged();
+        delayBlocks = maxTimeVariation_.delayBlocks;
+        futureBlocks = maxTimeVariation_.futureBlocks;
+        delaySeconds = maxTimeVariation_.delaySeconds;
+        futureSeconds = maxTimeVariation_.futureSeconds;
```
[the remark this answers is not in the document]
yahgwai: I think it's fine to have maxtimevariation as uint64s right? I've updated to that now

shotaronowhere:
The `modifier refundsGas(IGasRefunder gasRefunder)` in the [GasRefundEnabled](https://github.com/OffchainLabs/nitro-contracts/blob/05e311d8107e6f9ebd7ca66e256a7a494d3d66a5/src/libraries/IGasRefunder.sol#L26) abstract contract is unchanged, however it still calculates the gas refund assuming the calling contract is a proxy (copying calldata to memory). Since the sequencerInbox is no longer a proxy, the refundsGas modifier miscalculates. Underpayment is a problem, but overpayment isn't a big issue (this case),  however it's an unnecessary calculation (wasted gas).

I presume a refactor of the gasRefunder is planned given the eip 4844 support #96 , just highlighting the need.

src/bridge/SequencerInbox.sol
```diff
@@ -101,28 +104,62 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+        ) = maxTimeVariationInternal();
+
+        return
+            ISequencerInbox.MaxTimeVariation({
+                delayBlocks: delayBlocks_,
+                futureBlocks: futureBlocks_,
+                delaySeconds: delaySeconds_,
+                futureSeconds: futureSeconds_
+            });
+    }
```
Tristan-Wilson, line 146: To preserve backwards compatibility between versions can you make this return the old 4xuint256 struct? I think this is only called with eth_call so there is no need to save gas.

Tristan-Wilson (changes requested):

src/bridge/SequencerInbox.sol
gzeoneth: Hmm I suppose we can fix it like this https://github.com/OffchainLabs/nitro-contracts/commit/f3cfc60cd9cedb7b715dd3d4aad69bec43125e09 (not committed to this branch yet)

Tristan-Wilson: I think this will work since it's going back to the old signature https://github.com/OffchainLabs/nitro-contracts/pull/68/files/773e1ec74b404098f8eead510c727bfc471afad7#diff-f49bf9a48371b4b787c61abac8f18edfa65baa865c69e140b5a67d430070f245L81-L89

src/bridge/SequencerInbox.sol
```diff
@@ -56,7 +54,11 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
 
     IOwnable public rollup;
```
DZGoldman, line 55: why was rollup kept mutable? in anticipation for BOLD update or sthg?

src/bridge/ISequencerInbox.sol
```diff
@@ -52,6 +34,8 @@ interface ISequencerInbox is IDelayedMessageProvider {
     /// @dev a keyset was invalidated
     event InvalidateKeyset(bytes32 indexed keysetHash);
 
+    /// @notice The total number of delated messages read in the bridge
```
DZGoldman, line 37:
```suggestion
    /// @notice The total number of delayed messages read in the bridge
```

src/bridge/SequencerInbox.sol
yahgwai: Merged [f3cf](https://github.com/OffchainLabs/nitro-contracts/commit/f3cfc60cd9cedb7b715dd3d4aad69bec43125e09)

src/bridge/SequencerInbox.sol
yahgwai: Yeh, we may ship a new rollup and it would be nice not to have to update the seq inbox. As it happens the rollup address isnt used on the happy path, so making it mutable doesnt affect the gas cost of batch posting

src/bridge/AbsBridge.sol
```diff
@@ -55,8 +59,25 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
...
+        uint256 indexed batchSequenceNumber,
+        bytes32 indexed beforeAcc,
+        bytes32 indexed afterAcc,
+        bytes32 delayedAcc,
+        uint256 afterDelayedMessagesRead,
+        ICommon.TimeBounds timeBounds,
+        ICommon.BatchDataLocation dataLocation
+    );
+
+    function postUpgradeInit() external onlyDelegated onlyProxyOwner {
```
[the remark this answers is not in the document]
yahgwai: Yeh, I thinkn that makes sense, will update.

yahgwai:
> The `modifier refundsGas(IGasRefunder gasRefunder)` in the [GasRefundEnabled](https://github.com/OffchainLabs/nitro-contracts/blob/05e311d8107e6f9ebd7ca66e256a7a494d3d66a5/src/libraries/IGasRefunder.sol#L26) abstract contract is unchanged, however it still calculates the gas refund assuming the calling contract is a proxy (copying calldata to memory). Since the sequencerInbox is no longer a proxy, the refundsGas modifier miscalculates. Underpayment is a problem, but overpayment isn't a big issue (this case), however it's an unnecessary calculation (wasted gas).
> 
> I would either make the `GasRefundEnabled` contract delegate call aware, or avoid abstracting the gas refund logic as general purpose contract all together.
> 
> I presume a refactor of the gasRefunder is planned given the eip 4844 support #96 , just highlighting the need.

Great catch, updated the gasrefundenabled contract to be delegate call aware. 4844 gas refund changes will happen in a separate pr.

gzeoneth:
https://github.com/OffchainLabs/nitro-contracts/pull/110

PR #73, opened 2023-10-05
Add ERC20Bridge balance check

gvladika:
When withdraw-and-call is used in ERC20 based chains, it is important to make sure that extra call cannot be used to move escrowed native token in any way. We enforce it by doing native token balance check for bridge before/after the extra call is executed.

### DIFF
--- a/src/bridge/ERC20Bridge.sol
+++ b/src/bridge/ERC20Bridge.sol
@@ -15,6 +15,11 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
  * @title Staging ground for incoming and outgoing messages
  * @notice Unlike the standard Eth bridge, native token bridge escrows the custom ERC20 token which is
  * used as native currency on L2.
+ * @dev Fees are paid in this token. There are certain restrictions on the native token:
+ *       - The token can't be rebasing or have a transfer fee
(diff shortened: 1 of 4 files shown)


src/bridge/ERC20Bridge.sol
```diff
@@ -50,20 +50,30 @@ contract ERC20Bridge is AbsBridge, IERC20Bridge {
...
+        // if there's data do additional contract call. Make sure that call is not used to
+        // change bridge contract's balance of the native token
         if (data.length > 0) {
+            uint256 bridgeBalanceBefore = IERC20(_nativeToken).balanceOf(address(this));
+
             // solhint-disable-next-line avoid-low-level-calls
             (success, returnData) = to.call(data);
+
+            uint256 bridgeBalanceAfter = IERC20(_nativeToken).balanceOf(address(this));
+            if (bridgeBalanceAfter != bridgeBalanceBefore) {
```
gzeoneth, line 74: can be a bit looser? idk
```suggestion
            if (bridgeBalanceAfter < bridgeBalanceBefore) {
```

gvladika: yeah I don't see practical use case to increase balance, but it doesn't hurt. updated

gzeoneth (approved):
LGTM

PR #74, opened 2023-10-05
Add a batch poster manager for the sequencer inbox

yahgwai:
* At this point we should consider inheriting openzepplin.AccessControl as we now have three roles
  * Rollup owner
  * batch poster
  * batch poster manager


src/bridge/SequencerInbox.sol
```diff
@@ -54,6 +55,9 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
     bytes1 public constant DATA_AUTHENTICATED_FLAG = 0x40;
 
     IOwnable public rollup;
+    /// @notice The batch poster manager has the ability to change the batch poster addresses
+    ///         This enables the batch poster to do key rotation
+    address public batchPosterManager;
```
PlasmaPower, line 60: Should we add a setBatchPosterManager?

yahgwai: Yep, good catch! Updated

gzeoneth (dismissed):
LGTM

DZGoldman (approved):

gzeoneth (dismissed):
LGTM
Agree that inheriting OZ AccessControl can be nice, but that would mess up storage. One potential solution is we have a BridgeStorage base contract that inherit the current storage layout so it doesn't get shifted downward.

src/bridge/SequencerInbox.sol
```diff
@@ -469,11 +481,18 @@ contract SequencerInbox is GasRefundEnabled, ISequencerInbox {
     }
 
     /// @inheritdoc ISequencerInbox
-    function setIsSequencer(address addr, bool isSequencer_) external onlyRollupOwner {
+    function setIsSequencer(address addr, bool isSequencer_) external {
+        if (msg.sender != batchPosterManager) revert NotBatchPosterManager(msg.sender);
         isSequencer[addr] = isSequencer_;
```
DZGoldman, line 486: i think isSequencer isn't used on chain; shouldn't the batchPosterManager be calling setIsBatchPoster?

src/bridge/SequencerInbox.sol
```diff
@@ -469,11 +481,18 @@ contract SequencerInbox is GasRefundEnabled, ISequencerInbox {
...
         emit OwnerFunctionCalled(4);
```
DZGoldman, line 487: this (or if it changes, setIsBatchPoster) isn't an owner function anymore; should we change the event? (or at least add a comment?)

DZGoldman (changes requested):

src/bridge/SequencerInbox.sol
PlasmaPower: The batchPosterManager should control both. isSequencer is indeed used off-chain to determine who should be able to sign the feed, but who should be able to sign the feed should be controlled by who should be able to post batches as the latter is a strictly more powerful ability and they should be the same entity

DZGoldman:
as discussed: probably makes sense get https://github.com/OffchainLabs/nitro-contracts/pull/94 & https://github.com/OffchainLabs/nitro-contracts/pull/107 merged and then include batch poster management as a role on the Bridge

gzeoneth:
Pushed changes to base this on `4844-only`, some additional changes includes:
1. Rollup Owner is always a batch poster manager
2. Batch poster manager control both isSequencer and isBatchPoster
3. Revert to proxy pattern instead of immutable, batch poster manager can be set during initialization

src/bridge/SequencerInbox.sol
gzeoneth: hmm semantic, I prefer to keep the same event
```suggestion
        emit OwnerFunctionCalled(4); // Owner in this context can also be batch poster manager
```

yahgwai (commented):
LGTM

gzeoneth (approved):
LGTM

src/bridge/SequencerInbox.sol
```diff
@@ -99,6 +101,13 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
         _;
     }
 
+    modifier onlyRollupOwnerOrBatchPosterManager() {
+        if (msg.sender != rollup.owner() && msg.sender != batchPosterManager) {
+            revert NotBatchPosterManager(msg.sender);
```
DZGoldman, line 106: maybe `NotBatchPosterManagerOrRollupOwner`? (maybe fine as is)

DZGoldman (approved):

src/bridge/SequencerInbox.sol
gzeoneth: ikr but intentionally used the short form

PR #77, opened 2023-10-09
Fix the BridgeCreator.sol contract

gvladika:
Cherry-pick commits from https://github.com/tsite/nitro-contracts/tree/feature-orbit-bridge

Clean up the code and reduce deployment gas cost by moving contract creation outside of the contructor.

Move the sequencer inbox maxTimeVariation from memory to calldata.

Add DISABLE_VERIFICATION environment variable to disable contract verification.


gzeoneth (approved):
LGTM

PR #82, opened 2023-10-24
feat: outbox postUpgradeInit

gzeoneth:
Outbox L2ToL1Context struct is changed from
```solidity
    struct L2ToL1Context {
        uint128 l2Block; // slot 1
        uint128 l1Block; 
        uint128 timestamp; // slot 2
        bytes32 outputId; // slot 3
        address sender; // slot 4
    } 
```
to
```solidity
    struct L2ToL1Context {
        uint128 l2Block; // slot 1
        uint128 timestamp;
        bytes32 outputId; // slot 2
        address sender; // slot 3
        uint96 l1Block;
        uint256 withdrawalAmount; // slot 4
    }
```
This PR add `postUpgradeInit` to Outbox which is supposed to be called with upgradeAndCall to fix the default context.

### DIFF
--- a/src/bridge/AbsOutbox.sol
+++ b/src/bridge/AbsOutbox.sol
@@ -12,7 +12,8 @@ import {
     UnknownRoot,
     AlreadySpent,
     BridgeCallFailed,
-    HadZeroInit
+    HadZeroInit,
+    BadPostUpgradeInit
 } from "../libraries/Error.sol";
 import "./IBridge.sol";
 import "./IOutbox.sol";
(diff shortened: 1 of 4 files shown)


src/bridge/AbsOutbox.sol
```diff
@@ -76,6 +77,19 @@ abstract contract AbsOutbox is DelegateCallAware, IOutbox {
         rollup = address(_bridge.rollup());
     }
 
+    function postUpgradeInit() external onlyDelegated onlyProxyOwner {
```
gvladika, line 80: Since it only impacts existing ETH based rollups, we could put this implementation to `Outbox.sol` and revert if called on `ERC20Outbox.sol`?

gvladika (commented):
Lgtm. `OutboxWithoutOptTester` needs to implement new method to fix CI.

src/bridge/AbsOutbox.sol
gzeoneth: prefer to have it absoutbox, this is not specific to eth rollup, but rollup before 1.1.0

yahgwai (approved):

gvladika (approved):

PR #54, opened 2023-09-01
Add updateRollupAddress to bridge

amsanghi:
Copied it from https://github.com/OffchainLabs/bold/pull/340/

Needed for https://github.com/OffchainLabs/nitro/pull/1836

PR #48, opened 2023-08-02
Removed whenNotPaused from deposit withdrawal

yahgwai:
Do not merge.

Allows withdrawal from nitro contracts whilst paused.

PR #83, opened 2023-10-24
feat: allow rollup update

gzeoneth:
This PR included 2 changes that should only affect privileged methods

- [x] Allow rollup admin to update rollup addresses in Bridge contracts, and synchronizing those changes to Outbox, RollupEventInbox, and Sequencer Inbox. These changes are audited as part of BOLD in https://github.com/OffchainLabs/bold/pull/340 but for proxy admin instead of rollup admin. Allowing rollup admin to call those function make the upgrade executor being able to call those function easier (or otherwise would require a upgradeAndCall via the ProxyAdmin contract).

- [x] Introduce `whenNotPausedOrDeprecated` modifier, which allow validator to withdraw their stake when the rollup is deprecated by changing the rollup address in the bridge to something else

Superseding
* https://github.com/OffchainLabs/nitro-contracts/pull/54
* https://github.com/OffchainLabs/nitro-contracts/pull/48

### DIFF
--- a/src/rollup/RollupUserLogic.sol
+++ b/src/rollup/RollupUserLogic.sol
@@ -227,6 +227,11 @@ abstract contract AbsRollupUserLogic is
         stakeOnNode(msg.sender, latestNodeCreated());
     }
 
+    modifier whenNotPausedOrDeprecated() {
+        require(!paused() || address(bridge.rollup()) != address(this), "PAUSED_OR_NOT_DEPRECATED");
+        _;
+    }
+
     /**
      * @notice Refund a staker that is currently staked on or before the latest confirmed node
      * @dev Since a staker is initially placed in the latest confirmed node, if they don't move it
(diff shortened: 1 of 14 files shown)


src/rollup/RollupUserLogic.sol
```diff
@@ -227,14 +227,24 @@ abstract contract AbsRollupUserLogic is
         stakeOnNode(msg.sender, latestNodeCreated());
     }
 
+    modifier whenNotPausedOrDeprecated() {
```
yahgwai, line 230: Maybe move this modifier to the top of the contract

gvladika (dismissed):
lgtm, might be useful to emit RollupUpdated in Bridge

test/contract/arbRollup.spec.ts
```diff
@@ -789,6 +802,44 @@ describe('ArbRollup', () => {
     await rollup.confirmNextNode(challengerNode)
   })
 
+  it('allow force refund staker with pending node', async function () {
+    await rollupAdmin.connect(await impersonateAccount(upgradeExecutor)).pause()
```
yahgwai, line 806: missing .wait() - and on some other transactions below

yahgwai (approved):

test/contract/arbRollup.spec.ts
gzeoneth: that should not be necessary right? otherwise the test should be failing

yahgwai (approved):

test/contract/arbRollup.spec.ts
yahgwai: could just be getting lucky with timing. You might get intermittent failures with this

PR #86, opened 2023-10-25
make Bridge pausable

DZGoldman:
Currently pausing batch posting requires all batchPosters be removed in the sequencer inbox, pausing outbox execution requires all allowed outboxes be removed in the bridge, and pausing force inclusion isn't possible / requires a proxy upgrade. This makes those things possible more cleanly.

- Also removed acceptFundsFromOldBridge  while at it.
- AFAIK upgrading to inherit OZ's PausableUpgradeable would mess up the storage layout, so we just include the pause methods directly.


gvladika (commented):
There are also `enqueueDelayedMessage` entrypoints, defined in child contracts `ERC20Bridge` and `Bridge`, which can also be made pausable

src/bridge/AbsBridge.sol
```diff
@@ -66,6 +70,18 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
         }
         _;
     }
+    modifier whenNotPaused() {
```
yahgwai, line 73: I think it would be nice to have more granular pauses available in the bridge. Eg:
* pause withdrawal / outbox
* pause inbox / delayed messages (we already have this in the inbox contract so maybe we could remove it there and put it here instead)
* pause seq inbox messages

gzeoneth: As also mentioned here https://github.com/OffchainLabs/nitro-contracts/pull/74#pullrequestreview-1709957909
we can use OZ AccessControl by having a BridgeStorage base contract that inherit the current storage layout so it doesn't get shifted downward.

DZGoldman:
closing in favor of https://github.com/OffchainLabs/nitro-contracts/pull/107

PR #87, opened 2023-10-26
Move event and delayed inbox messages read to the bridge

yahgwai:
This makes state and events more consistent when sequencer inboxes are replaced.
* Events are emitted from the bridge
* Event signature changed due to addition of seq inbox address
* Requires a call to postUpgradeInit to initialise total delayed messages read


src/bridge/ICommon.sol
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021-2022, Offchain Labs, Inc.
+// For license information, see https://github.com/OffchainLabs/nitro-contracts/blob/main/LICENSE
+// SPDX-License-Identifier: BUSL-1.1
+
+// solhint-disable-next-line compiler-version
+pragma solidity >=0.6.9 <0.9.0;
+
+interface ICommon {
```
gzeoneth, line 8: the name is a bit confusing, why don't we just have these in IBridge?

src/bridge/AbsBridge.sol
```diff
@@ -55,8 +59,25 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
...
+        uint256 indexed batchSequenceNumber,
+        bytes32 indexed beforeAcc,
+        bytes32 indexed afterAcc,
+        bytes32 delayedAcc,
+        uint256 afterDelayedMessagesRead,
+        ICommon.TimeBounds timeBounds,
+        ICommon.BatchDataLocation dataLocation
+    );
+
+    function postUpgradeInit() external onlyDelegated onlyProxyOwner {
```
gzeoneth, line 76: let's mention this in PR description

src/bridge/AbsBridge.sol
```diff
@@ -55,8 +59,25 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
 
     uint256 public override sequencerReportedSubMessageCount;
 
+    uint256 public totalDelayedMessagesRead;
+
     address internal constant EMPTY_ACTIVEOUTBOX = address(type(uint160).max);
 
+    event SequencerBatchDelivered(
```
gzeoneth, line 66: we should document that the signature of this event will change due to the new packing of `TimeBounds`

gzeoneth (approved):
LGTM, some comments

src/bridge/ICommon.sol
yahgwai: removed and put in IBridge

src/bridge/AbsBridge.sol
yahgwai: Added desc on PR

gzeoneth (approved):
LGTM, should remove those dot file before merging

PR #88, opened 2023-10-26
Foundry storage check

gvladika:
Use 'forge inspect' to store and compare storage layouts as it is more readable then sol2uml


gzeoneth (approved):
LGTM

PR #90, opened 2023-10-27
feat: add sha256 preimage support to osp

gzeoneth:
* Was https://github.com/OffchainLabs/nitro-contracts/pull/50/


gvladika (approved):

gzeoneth:
LGTM, this should also be compatible with older version since the READ_PRE_IMAGE opcode does not take any argument (therefore argumentData should default to 0)

PR #92, opened 2023-10-30
chore: new storage check

gzeoneth:
This PR switch from the previous sol2uml storage to a foundry inspect based storage check. Was
* https://github.com/OffchainLabs/nitro-contracts/pull/88


gvladika (approved):
lgtm

PR #95, opened 2023-11-02
Added some basic seq inbox foundry tests

yahgwai:
Will add more in additional PRs

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -7,6 +7,11 @@ cache_path  = 'forge-cache'
 optimizer = true
 optimizer_runs = 20000
 via_ir = false
+# 0.8.21 has a bug which causes compilation failure in the tests
+# https://github.com/ethereum/solidity/issues/14430
+# however .22 is not shipped as default in foundry yet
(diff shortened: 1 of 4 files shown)


foundry.toml
```diff
@@ -7,6 +7,11 @@ cache_path  = 'forge-cache'
 optimizer = true
 optimizer_runs = 20000
 via_ir = false
+# 0.8.21 has a bug which causes compilation failure in the tests
+# https://github.com/ethereum/solidity/issues/14430
+# however .22 is not shipped as default in foundry yet
+# we can remove the line below when it is
+solc_version = '0.8.22'
```
gzeoneth, line 14: While the deployment uses the hardhat config, I think the foundry config should also matches the hardhat config. The hardhat config currently uses 0.8.9. 

Also note that solc above 0.8.19 introduced PUSH0 which Arbitrum One currently does not support, which can causes issue with orbit chains.
```suggestion
solc_version = '0.8.22'
evm_version = 'london'
```

gzeoneth (approved):

PR #96, opened 2023-11-03
4844 int opt

yahgwai:
Adds 4844 blob reading
* Reads data from blob
* Submits batch spending report of blobbasefee

Nitro changes
* New seq inbox function to submit blobs to
* Batch spending reports use a new gas price value


src/bridge/SequencerInbox.sol
```diff
@@ -307,37 +322,60 @@ contract SequencerInbox is GasRefundEnabled, ISequencerInbox {
...
+            if (dataHashes.length == 0) revert MissingDataHashes();
+            if (data.length != 1) revert InvalidBlobMetadata();
+            if (data[0] & DATA_BLOB_HEADER_FLAG == 0) revert InvalidBlobMetadata();
+
+            return (
+                keccak256(bytes.concat(header, data, abi.encodePacked(dataHashes))),
+                timeBounds
+            );
+        } else {
+            // the first byte is used to identify the type of batch data
```
PlasmaPower, line 361: We need to prevent the batch poster from posting a fake 4844 batch here by just specifying a header byte of `DATA_BLOB_HEADER_FLAG`. In addition, we need to prevent this from happening before the smart contract is upgraded. That's why the first 4844 PR used the `DATA_AUTHENTICATED_FLAG`

src/bridge/SequencerInbox.sol
```diff
@@ -307,37 +322,60 @@ contract SequencerInbox is GasRefundEnabled, ISequencerInbox {
...
         );
-        return (keccak256(header), timeBounds);
+        if (dataLocation == IBridge.BatchDataLocation.Blob) {
+            bytes32[] memory dataHashes = dataHashReader.getDataHashes();
+            if (dataHashes.length == 0) revert MissingDataHashes();
+            if (data.length != 1) revert InvalidBlobMetadata();
+            if (data[0] & DATA_BLOB_HEADER_FLAG == 0) revert InvalidBlobMetadata();
+
+            return (
+                keccak256(bytes.concat(header, data, abi.encodePacked(dataHashes))),
```
PlasmaPower, line 357: There's a risk here of if the batch poster appends a fake 32 byte data hash to their data. Right now the system can't tell the difference between the batch poster's data and the sequencer inbox generated data hashes, since they're right next to each other and not length delimited.

src/bridge/SequencerInbox.sol
```diff
@@ -245,48 +289,19 @@ contract SequencerInbox is GasRefundEnabled, ISequencerInbox {
...
+        addSequencerL2BatchImpl(
+            sequenceNumber,
+            dataHash,
+            timeBounds,
+            afterDelayedMessagesRead,
+            prevMessageCount,
+            newMessageCount,
+            IBridge.BatchDataLocation.SeparateBatchEvent
+        );
+        emit SequencerBatchData(sequenceNumber, data);
```
PlasmaPower, line 304: Did you look at how the first 4844 PR worked? It emitted the full data that L2 would read, including the data hashes and such, which simplifies the node's inbox reader a bit. It also means you don't need the new batch data location.

src/bridge/SequencerInbox.sol
yahgwai: Wouldnt that be prevented by the datahash reader erroring or returning no blobs?

src/bridge/SequencerInbox.sol
yahgwai: Perhaps I'm not understanding what you're saying, but the data is checked to be length 1 so that shouldnt be possible

src/bridge/SequencerInbox.sol
yahgwai: Yeh, I did see the PR. I need to take a look in the inbox reader I guess, but dont you have access to the full transaction data there, including the blob versioned hashes?

yahgwai: I assuming you mean this for the original PR: https://github.com/OffchainLabs/nitro-contracts/compare/develop...eip-4844-contracts

src/bridge/SequencerInbox.sol
PlasmaPower: Ah yeah sorry I missed that on my phone. We could consider requiring the data just be the blob header byte. Any extra header info could be put inside the data.

src/bridge/SequencerInbox.sol
PlasmaPower: Yeah I think so, cc @Tristan-Wilson would know better than me

src/bridge/SequencerInbox.sol
yahgwai: I think this would be prevented by looking at the batch data location in the SequencerBatchDelivered event.

PlasmaPower: The replay binary doesn't have access to that information, only the batch accumulator

src/bridge/SequencerInbox.sol
yahgwai: That is what we currently require, since we check length 1 and check the first byte is the blob header byte.

However I do agree that it's confusing to have the data here, when we're not really using it, I left it here for two reasons:
1. We may want to include extra 4844 related data in the future (maybe now?) such as the indexes of the blob versioned hashes we care about if the transaction sends to multiple entities.
2. It simplifies the internal representation to have a calldata object here

However, I think we can just do 1. if we need to, and 2. isnt important, so unless we think we want the granularity of the 4844 hash indices I'll remove it for now

src/bridge/SequencerInbox.sol
yahgwai: Hmm, ok, in that case I think the first PR was vulnerable too. I think we should check the first byte in all seq inbox functions to ensure the correct type of data (normal, da, blob, etc) is being provided in each case.

Prior to 4844 going live the data hash reader will not be able to return any hashes so the blob seq inbox function will revert. My assumption for nitro is that it would support 4844 blobs from the moment the EIP goes live or from the point that the seq inbox is upgraded. So I think with the above additional check we should be ok right?

PlasmaPower: The problem is that nitro doesn't know when the sequencer inbox contract is upgraded. If you're running a 4844 compatible nitro node (or replay binary) against an unupgraded sequencer inbox smart contract, and the batch poster maliciously posts a normal non-4844 batch with a first byte of `DATA_BLOB_HEADER_FLAG`, the replay binary doesn't have a good way of knowing that it's not a real 4844 batch and it shouldn't resolve the data hashes in it.

src/bridge/SequencerInbox.sol
Tristan-Wilson: Regarding the extra data I agree it doesn't hurt to have that facility. For the specific case of blob indices, I don't think the submitter of the tx knows the blob indices for the blobs that will be in the slot for the block their tx gets included in, since at the very least there can be multiple blob bearing txes included in that block.

yahgwai: The data block reader should only returns from this transaction, not from this block, which allows for index selection

PR #97, opened 2023-11-09
Deployment estimator script

gvladika:
Add script which estimates cost of deployment of RollupCreator

`yarn run estimate-factory --network arb1`


scripts/estimation.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const P = l2BaseFee
+  const L1P = l1BaseFee.mul(16)
+  const l1Size = gasEstimateComponents.gasEstimateForL1.mul(P).div(L1P)
+  const L1C = L1P.mul(l1Size)
+  const B = L1C.div(P)
+  const G = gasEstimateComponents.gasEstimate
+    .sub(gasEstimateComponents.gasEstimateForL1)
+    .add(B)
+  const TXFEES = P.mul(G)
```
gzeoneth, line 141: This simply reduce into `TXFEES = P * gasEstimateComponents.gasEstimate`, which seems to be a convoluted way of doing simple gas estimation. This should instead take an extra parameter to allow specifying target L1 gas price, and scale L1C based on that e.g. `L1C = L1P.mul(targetL1Fee).div(l1BaseFee).mul(l1Size)`

scripts/estimation.ts
```diff
@@ -0,0 +1,208 @@
...
+    gasEstimate: BigNumber
+    gasEstimateForL1: BigNumber
+    baseFee: BigNumber
+    l1BaseFeeEstimate: BigNumber
+  },
+  contractName: string
+) {
+  totalGas = totalGas.add(gasEstimateComponents.gasEstimate)
+  totalL1Gas = totalL1Gas.add(gasEstimateComponents.gasEstimateForL1)
+  totalL2Gas = totalL2Gas.add(
```
gzeoneth, line 125: instead of having these global vars, return the calculated scaled gas value and sum them within estimateAll

gzeoneth (changes requested):
Overall the variable naming is quite confusing, given 
1. gasEstimateComponents (gec)
2. targetL1GasPrice

the formula can simply be:
gec.baseFee * (gec.gasEstimate + gec.gasEstimateForL1 * (targetL1GasPrice / gec.l1BaseFeeEstimate) - 1)

PR #99, opened 2023-12-04
Add DepositMessageDelivered event

tsite:
This event allows bridges to efficiently query the deposit history for an address without requiring an external indexer.

### DIFF
--- a/src/bridge/AbsBridge.sol
+++ b/src/bridge/AbsBridge.sol
@@ -203,6 +203,7 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
             baseFeeL1,
             blockTimestamp
         );
+        emit MessageDeliveredV2(kind, sender);
         return count;
     }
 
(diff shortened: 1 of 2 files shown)


yahgwai:
Just noting that this adds around 1.1k gas to all delayed messages and batch spending reports

src/bridge/AbsBridge.sol
```diff
@@ -203,6 +203,7 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
             baseFeeL1,
             blockTimestamp
         );
+        emit MessageDeliveredV2(kind, sender);
```
gzeoneth, line 206: Might make more sense to have this event in the depositETH function so it does not add gas usage to other types of inbox messages. Can you elaborate more on the usecase of this feature?

gzeoneth (changes requested):

src/bridge/AbsBridge.sol
tsite: The main usecase is for a bridge application to show a user their deposits. Normally the MessageDelivered event is used for this, however since address is not indexed in this event, deposits for all users have to be queried which is much less efficient and requires more requests. Agreed that the event only needs to be emitted for deposit events - updated the logic to reflect that, which should save on gas.

PR #100, opened 2023-12-11
feat: normalize native token decimals

gzeoneth:
Was https://github.com/OffchainLabs/nitro-contracts/pull/65 but have decimal stored in bridge instead of inbox & outbox, also rebased the PR on main for fast-track release.

There is a scenario where custom fee token has number of decimals different than 18. That's not ideal because a lot of tooling assumes that native currency (on child chain) is using 18 decimals. For that reason we do a normalization, and de-normalization, when depositing and withdrawing native asset through native bridge.

Ie. let's say deposit amount is 752. 
 - if token has 16 decimals normalized amount which is going to be minted on child chain is 75200 
 - if token has 20 decimals normalized amount is 7
 - if token uses no decimals normalized amount is 752*10^18

When withdrawing, opposite process is done. There can be loss of precision, due to the fact that we have to round down amount to the nearest supported value.

We also add 2 requirements on the custom fee token in order to avoid overflow:
- fee token can use at most 36 decimals (enforced at inbox/outbox initialization)
- max amount that can be bridged is `type(uint256).max / 10**18` which is ~1.1*10^59 tokens

Also use safeApprove in ERC20Inbox.init to support token like USDT that does not return data on approve


src/bridge/ERC20Bridge.sol
```diff
@@ -24,15 +33,27 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
...
+        // store number of decimals used by native token
+        try ERC20(nativeToken_).decimals() returns (uint8 decimals) {
+            if (decimals > MAX_ALLOWED_NATIVE_TOKEN_DECIMALS) {
+                revert NativeTokenDecimalsTooLarge(decimals);
+            }
+            nativeTokenDecimals = decimals;
+        } catch {
+            // decimal is not part of the ERC20 spec
+            // assume it have 0 decimals if it does not have decimals() method
+            nativeTokenDecimals = 0;
```
yahgwai, line 55: I feel like we've discussed this before, but if the erc20 has no decimals might it not be better to assume 18 or revert and insist that a decimal value is provided in the constructor here. Do we have examples of tokens that dont use the decimals func? Do they actually use 0 decimals?

src/bridge/IERC20Bridge.sol
```diff
@@ -19,6 +19,8 @@ interface IERC20Bridge is IBridge {
      */
     function nativeToken() external view returns (address);
 
+    function nativeTokenDecimals() external view returns (uint8);
```
yahgwai, line 22: Should we add natspec?

src/bridge/ERC20Bridge.sol
```diff
@@ -24,15 +33,27 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 contract ERC20Bridge is AbsBridge, IERC20Bridge {
     using SafeERC20 for IERC20;
 
-    /// @inheritdoc IERC20Bridge
     address public nativeToken;
```
yahgwai, line 36: Why remove the inherit doc?

gzeoneth: 571303f

src/bridge/IERC20Bridge.sol
gzeoneth: 571303f

src/bridge/ERC20Bridge.sol
gzeoneth: 571303f

gvladika (approved):
lgtm

PR #101, opened 2023-12-14
fix: gas estimation issue

gzeoneth:


src/bridge/AbsInbox.sol
```diff
@@ -297,7 +297,9 @@ abstract contract AbsInbox is DelegateCallAware, PausableUpgradeable, IInboxBase
             revert GasLimitTooLarge();
         }
 
-        uint256 submissionFee = calculateRetryableSubmissionFee(data.length, block.basefee);
+        // block.basefee is used in calculateRetryableSubmissionFee when 0 is specified
+        // we use 0 instead of block.basefee to workaround a geth gas estimation quirk
```
yahgwai, line 301: I think we sgould improve the comment here by writing what the actual gas estimation quirk is, why this was a problem for us with the branching, and why setting 0 solves that.

gzeoneth: [e86195b](https://github.com/OffchainLabs/nitro-contracts/pull/101/commits/e86195ba2c71b6d8251481d7ac59566ab2a95e1c)

yahgwai (approved):

gzeoneth:
no-longer an issue

PR #102, opened 2023-12-22
Post audit fixes for native token decimals feature 

gvladika:
When chain uses custom fee token which has non-18 decimals, that means that amount of locked funds on the parent chain is different than the amount of funds that will be minted on the child chain.  This commit updates RollupCreator and DeployHelper (those contracts contain logic to deploy create2 deterministic factories to child chain via retrybles) so that amount of funds that needs to be locked is properly calculated.

Natspecs have been updated to make the denomination of retryable parameters more clear.

Also e2e tests have been refactored and improved to include more checks.


gzeoneth:
#100

gzeoneth (approved):
LGTM

PR #104, opened 2024-01-02
Add WAVM KZG preimage proof

PlasmaPower:
This adds a new type of preimage proof for Ethereum versioned hashes using the new EIP-4844 point evaluation precompile which will be introduced in Cancun along with blobs themselves. It's fine to deploy this before then though, because it'll just revert with "KZG_PRECOMPILE_MISSING" if the new preimage type is attempted to be proven before it's ready (and these preimages shouldn't even be attempted to be proven until 4844 is being utilized anyways).

This PR also requires that the WAVM ReadPreImage opcode be called with a preimage offset that's a multiple of 32, which was already what was done in practice before this.

### DIFF
--- a/src/osp/OneStepProverHostIo.sol
+++ b/src/osp/OneStepProverHostIo.sol
@@ -106,6 +106,25 @@ contract OneStepProverHostIo is IOneStepProver {
     uint256 constant PRIMITIVE_ROOT_OF_UNITY =
         10238227357739495823651030575849232062558860180284477541189508159991286009131;
 
+    // Computes b**e % m
+    // Really pure but the Solidity compiler sees the staticcall and requires view
+    function modExp256(uint256 b, uint256 e, uint256 m) internal view returns (uint256) {
+        bytes memory modExpInput = abi.encode(
+            32,
+            32,
+            32,
+            b,
+            e,
+            m
+        );
+        (bool modexpSuccess, bytes memory modExpOutput) = address(0x05).staticcall(
+            modExpInput
+        );
+        require(modexpSuccess, "MODEXP_FAILED");
+        require(modExpOutput.length == 32, "MODEXP_WRONG_LENGTH");
+        return uint256(bytes32(modExpOutput));
+    }
+
     function executeReadPreImage(
         ExecutionContext calldata,
         Machine memory mach,
@@ -195,21 +214,11 @@ contract OneStepProverHostIo is IOneStepProver {
 
                 uint256 rootOfUnityPower = (1 << 32) / fieldElementsPerBlob;
                 rootOfUnityPower *= bitReversedIndex;
-                bytes memory modExpInput = abi.encode(
-                    32,
-                    32,
-                    32,
+                uint256 rootOfUnity = modExp256(
                     PRIMITIVE_ROOT_OF_UNITY,
                     rootOfUnityPower,
                     BLS_MODULUS
                 );
-                (bool modexpSuccess, bytes memory modExpOutput) = address(0x05).staticcall(
-                    modExpInput
-                );
-                require(modexpSuccess, "MODEXP_FAILED");
-
-                uint256 rootOfUnity = abi.decode(modExpOutput, (uint256));
-
                 require(bytes32(kzgProof[32:64]) == bytes32(rootOfUnity), "KZG_PROOF_WRONG_Z");
 
                 extracted = kzgProof[64:96];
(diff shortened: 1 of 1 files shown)


src/osp/OneStepProverHostIo.sol
```diff
@@ -158,6 +177,47 @@ contract OneStepProverHostIo is IOneStepProver {
...
+                    }
+                    tmp >>= 1;
+                }
+
+                uint256 rootOfUnityPower = (1 << 32) / fieldElementsPerBlob;
+                rootOfUnityPower *= bitReversedIndex;
+                uint256 rootOfUnity = modExp256(
+                    PRIMITIVE_ROOT_OF_UNITY,
+                    rootOfUnityPower,
+                    BLS_MODULUS
```
yahgwai, line 215: Why do we use a hardcoded BLS_MODULUS instead of the one returned from the precompile?

src/osp/OneStepProverHostIo.sol
```diff
@@ -158,6 +177,47 @@ contract OneStepProverHostIo is IOneStepProver {
...
+                    bitReversedIndex <<= 1;
+                    if (tmp & 1 == 1) {
+                        bitReversedIndex |= 1;
+                    }
+                    tmp >>= 1;
+                }
+
+                uint256 rootOfUnityPower = (1 << 32) / fieldElementsPerBlob;
+                rootOfUnityPower *= bitReversedIndex;
+                uint256 rootOfUnity = modExp256(
```
yahgwai, line 212: Would this be better named z? Maybe we should include some comments explaining that the precompile expects the caller to raise the root of unity to the power of the index before calling?

src/osp/OneStepProverHostIo.sol
```diff
@@ -101,16 +101,35 @@ contract OneStepProverHostIo is IOneStepProver {
         state.u64Vals[idx] = val;
     }
 
+    uint256 constant BLS_MODULUS =
+        52435875175126190479447740508185965837690552500527637822603658699938581184513;
+    uint256 constant PRIMITIVE_ROOT_OF_UNITY =
```
yahgwai, line 106: Could we provide this with the proof data, or have it be part of the machine, to avoid needing to upgrade the contracts in future if this param changes for a different version? They talk about enabling that on the EIP, but I'm not sure it's worth it vs the simplicity of hardcoding it here. What do you think?

src/osp/OneStepProverHostIo.sol
```diff
@@ -158,6 +177,47 @@ contract OneStepProverHostIo is IOneStepProver {
...
+
+            require(bytes32(kzgProof[:32]) == leafContents, "KZG_PROOF_WRONG_HASH");
+
+            (bool success, bytes memory kzgParams) = address(0x0A).staticcall(kzgProof);
+            require(success, "INVALID_KZG_PROOF");
+            require(kzgParams.length > 0, "KZG_PRECOMPILE_MISSING");
+            (uint256 fieldElementsPerBlob, ) = abi.decode(kzgParams, (uint256, uint256));
+
+            // If preimageOffset is greater than the blob size, leave extracted empty and call it here.
+            if (preimageOffset < fieldElementsPerBlob * 32) {
```
yahgwai, line 195: Why do we return empty here instead of reverting?

src/osp/OneStepProverHostIo.sol
PlasmaPower: Hmm.. I think if the user supplies the root of unity, and we validate that `modExp256(rootOfUnity, fieldElementsPerBlob, blsModulus) == 1` we don't need to hardcode the root of unity, but personally I'd be more worried about that than just hardcoding it and upgrading it if necessary.

src/osp/OneStepProverHostIo.sol
PlasmaPower: Fixed, I think this code was originally before the precompile call which is why it was hardcoded

src/osp/OneStepProverHostIo.sol
PlasmaPower: Yeah I don't think rootOfUnity is correct. I've renamed this to z and added some comments.

src/osp/OneStepProverHostIo.sol
PlasmaPower: This is according to the API of the ReadPreImage opcode. If the offset requested is past the end of the preimage, an empty byte array is returned. That's so that the WASM can figure out the length of the preimage. The replay binary keeps calling ReadPreImage at 32 byte increments until it reaches a preimage of size <32. Theoretically, we could put the machine in the error state if preimageOffset > fieldElementsPerBlob * 32, but the API seems simpler to me if we just return an empty byte array in that case.

yahgwai (approved):

PR #107, opened 2024-01-10
Bridge: add pausable outbox execution, inbox updating, and delayed msg enqueueing

DZGoldman:
the pause delayed msg enqueuing role supplants the pausable inbox if we want to remove that

### DIFF
--- a/src/bridge/BridgePausable.sol
+++ b/src/bridge/BridgePausable.sol
@@ -1,0 +1,133 @@
+// Copyright 2021-2022, Offchain Labs, Inc.
+// For license information, see https://github.com/nitro/blob/master/LICENSE
+// SPDX-License-Identifier: BUSL-1.1
+import "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol";
+import {
+    OutboxExecutionPaused,
+    OutboxExecutionNotPaused,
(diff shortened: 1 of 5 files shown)


DZGoldman:
hm, so currently "inbox updating" = both batch posting and force-including; maybe we'd want to make those separately pausable?

DZGoldman:
^ maybe we remove PausableUpgradable from the inbox inheritance and add an empty storage slot for _paused?

src/bridge/BridgePausable.sol
```diff
@@ -0,0 +1,133 @@
...
+        if (!_delayedMessageEnqueuePaused) {
+            revert DelayedMessagesEnqueueNotPaused();
+        }
+        _;
+    }
+
+    function _grantAllPauseRolesTo(address owner) internal {
+        _grantRole(DEFAULT_ADMIN_ROLE, owner);
+        _grantRole(PAUSE_OUTBOX_EXECUTION_ROLE, owner);
+        _grantRole(UNPAUSE_OUTBOX_EXECUTION_ROLE, owner);
```
DZGoldman, line 79: thinking maybe the separate pause/unpause roles aren't necessary (and if we wanted to split up pause/unpause affordances we could do so via smart contract wallet)

PR #109, opened 2024-01-17
feat: 4844 implementation

gzeoneth:
Was #96, added [3b59719](https://github.com/OffchainLabs/nitro-contracts/pull/109/commits/3b59719ae77df665d54d4f7afcb41bd7b8788c06) to remove certain sequencer inbox optimizations.


src/bridge/SequencerInbox.sol
```diff
@@ -356,24 +490,136 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+            keccak256(bytes.concat(header, DATA_BLOB_HEADER_FLAG, abi.encodePacked(dataHashes))),
+            timeBounds
+        );
+    }
+
+    /// @dev   Submit a batch spending report message so that the batch poster can be reimbursed on the rollup
+    /// @param dataHash The hash of the message the spending report is being submitted for
+    /// @param seqMessageIndex The index of the message to submit the spending report for
+    /// @param gasPrice The gas price that was paid for the data (standard gas or data gas)
+    function submitBatchSpendingReport(
```
yahgwai, line 584: This is now almost duplicate code of what we have in addSequencerL2BatchImpl, should we call this instead? Or are you wanting to do that refactor in a later PR? if so we should leave a comment addSequencerL2BatchImpl explaining why we havent done it now

src/libraries/Error.sol
```diff
@@ -176,5 +176,26 @@ error AlreadyValidDASKeyset(bytes32);
...
+error MissingDataHashes();
+
+/// @dev Thrown when the data blob meta data is invalid
+error InvalidBlobMetadata();
+
 /// @dev Thrown when rollup is not updated with updateRollupAddress
 error RollupNotChanged();
+
+/// @dev Thrown when reading the total delayed messages and getting 0 when a non zero value is expected
+error EmptyDelayedMessagesRead();
```
yahgwai, line 195: Not used in this PR

src/rollup/RollupCreator.sol
```diff
@@ -102,29 +102,35 @@ contract RollupCreator is Ownable {
...
     {
-        // Make sure the immutable maxDataSize is as expected
-        (, ISequencerInbox ethSequencerInbox, IInboxBase ethInbox, , ) = bridgeCreator
-            .ethBasedTemplates();
-        require(
-            deployParams.maxDataSize == ethSequencerInbox.maxDataSize(),
-            "SI_MAX_DATA_SIZE_MISMATCH"
-        );
-        require(deployParams.maxDataSize == ethInbox.maxDataSize(), "I_MAX_DATA_SIZE_MISMATCH");
+        {
```
yahgwai, line 113: Why this change?

test/foundry/SequencerInbox.t.sol
```diff
@@ -144,7 +150,7 @@ contract SequencerInboxTest is Test {
         address delayedInboxSender = address(140);
         uint8 delayedInboxKind = 3;
         bytes32 messageDataHash = RAND.Bytes32();
-        bytes memory data = hex"a4567890"; // CHRIS: TODO: bigger data;
+        bytes memory data = hex"00a4567890"; // CHRIS: TODO: bigger data; // 00 is BROTLI_MESSAGE_HEADER_FLAG
```
yahgwai, line 153: We should address these CHRIS:TODO:. I'll add a commit

yahgwai (commented):
One downside here is that we've lost the commit history of the old PR

gzeoneth:
> One downside here is that we've lost the commit history of the old PR

I know, I can instead squash this on top of #96 to "revert" the immutable sequencer inbox changes. Do you think that sounds better?

src/bridge/SequencerInbox.sol
```diff
@@ -52,11 +59,27 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+    /// @inheritdoc ISequencerInbox
+    bytes1 public constant BROTLI_MESSAGE_HEADER_FLAG = 0x00;
+
+    /// @inheritdoc ISequencerInbox
+    bytes1 public constant ZERO_HEAVY_MESSAGE_HEADER_FLAG = 0x20;
 
     IOwnable public rollup;
     mapping(address => bool) public isBatchPoster;
-    ISequencerInbox.MaxTimeVariation public maxTimeVariation;
+    // see ISequencerInbox.MaxTimeVariation
```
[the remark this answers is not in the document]
gzeoneth: This is an internal change where the api can be backward compatible, tho it changed storage so I pushed a new commit to fix it https://github.com/OffchainLabs/nitro-contracts/pull/109/commits/6404ef1ad17dc03892305ed52da0f9cd3a156e69

src/libraries/Error.sol
gzeoneth: Removed.

src/rollup/RollupCreator.sol
gzeoneth: Stack size with the new `dataHashReader ` in `RollupDeploymentParams `

src/bridge/SequencerInbox.sol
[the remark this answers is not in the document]
yahgwai: It looks like we've introduced both the incompatibility and the fix for it in this change though. We could just do nothing with the maxtimevariation instead right?

src/rollup/RollupCreator.sol
yahgwai: Weird, I would have expected there just to be a mem pointer to that struct on the stack, not all it's contents. That datahashreader property isnt used in this function so why is it on the stack?

yahgwai:
> > One downside here is that we've lost the commit history of the old PR
> 
> I know, I can instead squash this on top of #96 to "revert" the immutable sequencer inbox changes. Do you think that sounds better? That way the commit history will preserves but a bit messier.

Yeh, that may be better, but only if it's not too crazy messy lol

gzeoneth:
Sorry for the force-push spam, should be good now.

src/bridge/SequencerInbox.sol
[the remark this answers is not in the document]
gzeoneth: We can keep it minimal, but I just feel like this seems to be something small enough (with meaningful gas saving) that can be included.

src/bridge/SequencerInbox.sol
gzeoneth: This would be a breaking change if we use tx.origin, and they are not getting the refund immediately anyway (it would be on L2) so I rather to have to current behavior.

src/bridge/SequencerInbox.sol
```diff
@@ -356,24 +538,139 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+        internal
+        view
+        returns (
+            bytes32,
+            IBridge.TimeBounds memory,
+            uint256
+        )
+    {
+        bytes32[] memory dataHashes = reader4844.getDataHashes();
+        if (dataHashes.length == 0) revert MissingDataHashes();
```
PlasmaPower, line 620: No change necessary, I just wanted to point out that this is a great check and we should keep it. Specifically, I'm worried that some RPC providers that aren't ready for 4844 yet might drop the blobVersionedHashes field from eth_estimateGas and would otherwise return too low of an estimate. Reverting here prevents that from happening.

src/bridge/SequencerInbox.sol
```diff
@@ -260,16 +389,81 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
             afterAcc,
             delayedAcc,
             totalDelayedMessagesRead,
             timeBounds_,
-            BatchDataLocation.TxInput
+            IBridge.BatchDataLocation.TxInput
+        );
+    }
+
+    function addSequencerL2BatchFromBlob(
```
PlasmaPower, line 409: nit: I think this should be "FromBlobs" plural since it can use multiple blobs

gzeoneth: Thanks! merged

yahgwai (dismissed):

yahgwai (dismissed):

src/bridge/SequencerInbox.sol
```diff
@@ -260,17 +392,81 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+
+        // blobs are currently not supported on host arbitrum chains, when support is added it may
+        // consume gas in a different way to L1, so explicitly block host arb chains so that if support for blobs
+        // on arb is added it will need to explicitly turned on in the sequencer inbox
+        if (hostChainIsArbitrum) revert DataBlobsNotSupported();
+
+        // submit a batch spending report to refund the entity that produced the blob batch data
+        uint256 blobGas = 0;
+        if (block.basefee > 0) {
+            blobGas = blobCost / block.basefee;
```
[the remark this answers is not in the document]
gzeoneth: 86b2b09

yahgwai (approved):

PR #111, opened 2024-01-17
Blob data gas refunds

yahgwai:
Now refunds if there were blobs on the transaction


src/bridge/GasRefunder.sol
```diff
@@ -216,11 +222,20 @@ contract GasRefunder is IGasRefunder, Ownable {
         uint256 maxRefundeeBalance = commonParams.maxRefundeeBalance;
         uint256 maxSingleGasUsage = commonParams.maxSingleGasUsage;
 
+        // add any cost for 4844 data, the data hash reader throws an error prior to 4844 being activated
+        try dataHashReader.getDataHashes() returns (bytes32[] memory dataHashes) {
```
PlasmaPower, line 226: I'd put this in the gasUsed calculation of the caller if tx.origin == msg.sender, because as-is a validator might be able to make many calls to their own wallet in one tx while supplying a blob and get refunded for each of them.

src/bridge/GasRefunder.sol
```diff
@@ -216,11 +222,20 @@ contract GasRefunder is IGasRefunder, Ownable {
...
+            if (dataHashes.length != 0) {
+                uint256 blobBasefee = blobBasefeeReader.getBlobBaseFee();
+                gasUsed += dataHashes.length * gasPerBlob * blobBasefee;
```
PlasmaPower, line 229: Shouldn't we `/ block.basefee` since gasUsed is measured in gas not wei?

yahgwai: yes, good catch!

src/bridge/GasRefunder.sol
yahgwai: Yeh, defo safer

PlasmaPower (approved):
LGTM. I'd consider unifying the blob basefee reader contract with the data hash reader contract but that's mostly separate.

gzeoneth (approved):
LGTM

gzeoneth:
Sorry for the force-push spam, should be good now.

src/libraries/GasRefundEnabled.sol
gzeoneth, line 1: need to update gas refunder in addSequencerL2BatchFromOrigin

gzeoneth (changes requested):

src/bridge/SequencerInbox.sol
```diff
@@ -38,10 +38,11 @@ import "../rollup/IRollupLogic.sol";
 import "./Messages.sol";
 import "../precompiles/ArbGasInfo.sol";
 import "../precompiles/ArbSys.sol";
+import "../libraries/I4844Readers.sol";
 
 import {L1MessageType_batchPostingReport} from "../libraries/MessageTypes.sol";
-import {GasRefundEnabled, IGasRefunder} from "../libraries/IGasRefunder.sol";
-import "../libraries/DelegateCallAware.sol";
```
gzeoneth, line 44: also this seems to be removed by accident

src/libraries/GasRefundEnabled.sol
gzeoneth: I tried to resolve but hitting stack issue, maybe we should have 2 different modifier instead?
refundsGas(x) and refundsGas(x,y,z)

src/bridge/SequencerInbox.sol
gzeoneth: [e3cc721](https://github.com/OffchainLabs/nitro-contracts/pull/111/commits/e3cc7214d4f734d06d3476798e641ae68bbf5ce8)

src/libraries/GasRefundEnabled.sol
shotaronowhere: or simply refundsGas(x) and refundsGasWithBlob(x)

shotaronowhere (changes requested):
As gzeon mentioned, gas refunder in `addSequencerL2BatchFromOrigin` should be updated, but there is a stack too deep error with extra arguments for the modifier.

Plausibly `refundsGas(x,y,z)` could be refactored to `refundsGas(x, bool refundBlob)` or simply `refundsGas(x)`, `refundsGasWithBlob(x)`

src/libraries/GasRefundEnabled.sol
yahgwai: ah yeh, I think it's missing because we rebased this PR onto this branch that has two variants of that func, whereas the old PR only had one.

PR #114, opened 2024-01-19
Include blob cost in extraGas field of batch posting reports

PlasmaPower:
Instead of adding a new field for the blob basefee, this PR adds the blob cost into the existing extraGas field.

The main disadvantage of this PR is that ArbOS will think that the sequencer paid for the blob hash list data as tx data, but I think that should be a small error.


gzeoneth (approved):
LGTM

PR #115, opened 2024-01-19
Add safety checks to Reader4844.yul

PlasmaPower:
Specifically, the contract will now revert if passed callvalue or supplied an unknown selector.


gzeoneth (approved):
LGTM

PR #118, opened 2024-01-23
Use 0x50 as the header byte for 4844 batches

PlasmaPower:
This leaves other possibilities open for the DATA_AUTHENTICATED_FLAG.

Also renames `addSequencerL2BatchFromBlob` to `addSequencerL2BatchFromBlobs` (plural) since it can use multiple blobs.


gzeoneth (approved):
LGTM

PR #121, opened 2024-01-24
misc: spending report logic and interface

gzeoneth:


src/bridge/SequencerInbox.sol
```diff
@@ -429,11 +429,11 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
         if (hostChainIsArbitrum) revert DataBlobsNotSupported();
 
         // submit a batch spending report to refund the entity that produced the blob batch data
-        uint256 blobGas = 0;
-        if (block.basefee > 0) {
-            blobGas = blobCost / block.basefee;
+        // same as using calldata, we only submit spending report if the caller is the origin of the tx
```
yahgwai, line 432: I think we should include more comments about why we do this

yahgwai (approved):

src/bridge/SequencerInbox.sol
gzeoneth: https://github.com/OffchainLabs/nitro-contracts/blob/22dd3a7cb7106cc397c6a874399f3078eaf8c3c0/src/bridge/SequencerInbox.sol#L615-L618

PR #142, opened 2024-02-12
fix: struct change

gzeoneth:
Revert a struct change that affected the 4bytes of `initialize` and `setMaxTimeVariation`, and anything that uses the Config struct e.g. `RollupAdminLogic`, `BridgeCreator`, `RollupCreator`, `RollupProxy`, etc.

The affected functions are either privileged or only called on deploy. All storages are correctly handled.


gvladika (dismissed):
lgtm

gvladika (approved):

PR #148, opened 2024-02-22
Add Slither to CI

gvladika:
This PR adds Slither to CI. Slither will run on every PR. It is configured so that job fails if there are any findings of >=medium severity found. Findings are uploaded to: https://github.com/OffchainLabs/nitro-contracts/security/code-scanning and can be filtered by PR number.

Initial false positive findings are added to exception list (`slither.db.json`) by running `slither . --triage-mode`. If there is any new false positive it can be added to the existing list in the same way.


gzeoneth (approved):
LGTM

gvladika:
> need to fix a new slither test

Updated triage db

gzeoneth (approved):

PR #151, opened 2024-02-26
chore: remove hardhat patch and fix foundry config

gzeoneth:


gzeoneth:
merged in https://github.com/OffchainLabs/nitro-contracts/pull/152, will ship with next patch release

PR #156, opened 2024-03-04
Orbit versioning

gvladika:
Add script which checks version of deployed Orbit bridge contracts.

Script requires `INBOX_ADDRESS` as the inout in `.env`. 

Then it can be executed like this, providing the name of parent network (needs to match hardhat config definition)

```
 yarn orbit:contracts:version --network arb1
```


gzeoneth:
https://github.com/OffchainLabs/orbit-actions/pull/7

PR #159, opened 2024-03-15
Add function to test if arbBlockHash on pending block returns latest

ganeshvanahalli:


src/mocks/PendingBlkTimeAndNrAdvanceCheck.sol
```diff
@@ -19,4 +19,12 @@ contract PendingBlkTimeAndNrAdvanceCheck {
         require(block.timestamp > deployedAt, "Time didn't advance");
         require(ArbSys(address(100)).arbBlockNumber() > deployedAtBlock, "Block didn't advance");
     }
+
+    function checkArbBlockHashReturnsLatest(bytes32 expected) external {
+        bytes32 gotBlockHash = ArbSys(address(100)).arbBlockHash(
+            ArbSys(address(100)).arbBlockNumber() - 1
```
PlasmaPower, line 25: IIRC to simplify this and the constructor, you can declare a constant in the contract block next to the immutables before the functions:
```
ArbSys constant arbSys = ArbSys(address(100));
```
and then you can just do `arbSys.arbBlockNumber()`

I might have misremembered that though

ganeshvanahalli: Cool, thanks for this!

gzeoneth (approved):
LGTM

PR #160, opened 2024-03-18
feat: delay buffer

shotaronowhere:
Introduces a delay buffer to limit repeated sequencer censorship enabling optimistic mechanisms (eg interactive L3 fraud proofs) on L2s without unreasonable delay. The delay buffer is represented by a struct stored in the SequencerInbox, and managed with the internal DelayBuffer library.

There are 2 ways to post batches. 

1. No Proof

No changes to existing batch posting function selectors. This method of batch posting is permissible when the sequencer inbox is in a synced state OR if no new delayed messages are read. In the former case, when the sequencer is including delay messages without unexpected delay, there is a minimum amount of time during which no unexpected delays are possible, and hence the sequencer can post batches without additional delay buffer related logic. Eg. if the delay threshold is 2 hours, and the sequencer proves the most recent sequenced delayed message is delayed 30 min, then the margin of 1 hr 30 min defines a sync period during which it is not possible to have unexpectedly delayed messages. In the later case, when no new delayed messages are read, no buffer state updates can be calculated since the elapsed time in the delayed message queue is zero.

2. Delay Proof

When the sequencer reads atleast one new delayed message, it can include a pre-image of the first delayed message read in the batch of the following form
```
        bytes32 beforeDelayedAcc,
        Messages.Message calldata delayedMessage
```
in order to calculate any delays and apply buffer updates as well as establish a sync state.

---

The batch posting strategy should include delayProofs if new delayed messages are read, otherwise, the original batch posting method without delay proof is sufficient.


src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,393 @@
...
+        roundOff = (elapsed + roundOff) % period;
+        if (maxBuffer - buffer > replenishAmount) {
+            buffer += replenishAmount;
+        } else {
+            // saturate
+            buffer = maxBuffer;
+            roundOff = 0;
+        }
+        return (buffer, roundOff);
+    }
```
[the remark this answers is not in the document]
shotaronowhere: calculation round-off is explicitly handled and managed.

godzillaba (commented):
so far it looks good, i need to explore the non happy `sync` path further though. 

i just have some very small suggestions for now

src/bridge/SequencerInbox.sol
```diff
@@ -363,55 +398,158 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
         // solhint-disable-next-line avoid-tx-origin
         if (msg.sender != tx.origin) revert NotOrigin();
         if (!isBatchPoster[msg.sender]) revert NotBatchPoster();
-        (bytes32 dataHash, IBridge.TimeBounds memory timeBounds) = formCallDataHash(
+        if (isDelayBufferable && !isSynced()) {
+            if (afterDelayedMessagesRead > totalDelayedMessagesRead) {
+                revert DelayProofRequired();
+            }
+            // no proof required if no new delayed messages are read since no buffer updates can be applied
+        }
```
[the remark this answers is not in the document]
shotaronowhere: hmmm true, I think for readability it's best as is, but I don't feel strongly about this. Could also be abstracted in an internal function

src/bridge/SequencerInbox.sol
```diff
@@ -363,55 +398,158 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
-        emit SequencerBatchDelivered(
-            seqMessageIndex,
-            beforeAcc,
-            afterAcc,
-            delayedAcc,
-            totalDelayedMessagesRead,
-            timeBounds_,
-            IBridge.BatchDataLocation.TxInput
+    /// @inheritdoc ISequencerInbox
+    function addSequencerL2BatchFromBlobs(
```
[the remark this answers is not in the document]
shotaronowhere: I refactored this by creating a struct for the extra proof.

function addSequencerL2BatchFromBlobs(...)
function addSequencerL2BatchFromBlobs(..., delayProof)
function addSequencerL2BatchFromBlobs(..., syncProof)

I could also update the function name as well, but I thought this approach is readable and clean.

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,340 @@
...
+
+    /// @notice Conditionally depletes or replenishes the delay buffer
+    /// @notice Decrements or replenishes the delay buffer conditionally
```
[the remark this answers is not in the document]
shotaronowhere: fixed

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,347 @@
...
+        uint64 threshold,
+        uint64 buffer
+    ) internal pure returns (uint64) {
+        uint64 elapsed = end > start ? end - start : 0;
+        uint64 unexpectedDelay = delay > threshold ? delay - threshold : 0;
+        uint64 decrease = unexpectedDelay > elapsed ? elapsed : unexpectedDelay;
+        // decrease the buffer saturating at zero
+        buffer = buffer > decrease ? buffer - decrease : 0;
+        // saturate at threshold
+        buffer = buffer > threshold ? buffer : threshold;
```
godzillaba, line 176: i don't really understand saturating at the threshold

godzillaba (approved):

godzillaba:
looks good!

src/bridge/DelayBuffer.sol
shotaronowhere: what would happen if the buffer is allowed to deplete to zero?

L1->L2 messages could be force included immediately. Force inclusion increments the sequencerNumber, so the sequencer attempting to include a batch may have their batches revert due to a sequencerNumber mismatch. The sequencer would need to post batches setting the sequencerNumber to type(uint256).max to bypass the check.

Further, the sequencer could not give out any preconfirmations when it recovers (back online after outage) since there is no delay on force inclusions and therefore L2 reorgs are a concern. The sequencer would be required to sequence exclusively delayed messages until the buffer replenishes sufficiently (could be hours to days). Arguably this is a good feature.

----

Alternatively, as written, the buffer minimum is the threshold (perhaps 30 min, or a couple hours). Then when the sequencer recovers is has some margin to issue preconfirmations immediately. This also means that L1->L2 messages would be required to wait atleast the threshold at all times.

For the requirements of reasonable challenge times for L3 fraud proof games, this means that 50 steps of bisection would require 50*threshold more time compared to the L2 fraud proof game on L1. So for reasonable threshold ranges (30 min - 2 hours), than implies an additional challenge time for L3s of a few days.

---

What do you think? Saturation at the threshold allows the sequencer to recover "smoothly". The trade off is minimum force inclusion time which doesn't have large impact on the delays of L3 fraud proof games. On the otherhand, allowing the buffer to drop to zero would mean minimal delay for L3 fraud proofs, but somewhat more awkward recovery for the Sequencer.

src/bridge/SequencerInbox.sol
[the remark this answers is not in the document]
godzillaba: i don't feel strongly either, seems good as is

src/rollup/BridgeCreator.sol
```diff
@@ -21,6 +21,8 @@ import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 contract BridgeCreator is Ownable {
     BridgeContracts public ethBasedTemplates;
     BridgeContracts public erc20BasedTemplates;
+    BridgeContracts public ethBasedDelayBufferableTemplates;
+    BridgeContracts public erc20BasedDelayBufferableTemplates;
```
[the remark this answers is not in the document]
gzeoneth: second this
sequencer inbox is the same for eth/erc20, so we can remove it from `BridgeContracts` and have additional var to store sequencerInbox and sequencerInboxBufferable template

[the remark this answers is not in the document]
gvladika: We introduced `isUsingFeeToken` immutable to seq inbox in v1.2.0, so I think we'd need 4 seq inbox template combos (fee token x bufferable)

src/bridge/SequencerInbox.sol
```diff
@@ -712,6 +946,67 @@ contract SequencerInbox is DelegateCallAware, GasRefundEnabled, ISequencerInbox
...
+        if (!DelayBuffer.isValidBufferConfig(bufferConfig_)) {
+            revert BadBufferConfig();
+        }
+
+        uint64 syncBlockNumber = buffer.syncExpiryBlockNumber > 0 &&
+            buffer.syncExpiryBlockNumber != type(uint64).max
+            ? buffer.syncExpiryBlockNumber - bufferConfig.thresholdBlocks
+            : 0;
+        uint64 syncTimestamp = buffer.syncExpiryTimestamp > 0 &&
+            buffer.syncExpiryBlockNumber != type(uint64).max
```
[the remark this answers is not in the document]
shotaronowhere: when the threshold changes, the sync expiry calculation must be updated since it may no longer be valid (threshold dropped from 10 hours to 1 hour, the outdated sync expiry may be valid for the next 9hr 30min, but we should adjust this to a 30 min validity)

essentially this calculation recovers the original timestamp / blocknumber used to calculate the sync validity in the first place.

src/rollup/BridgeCreator.sol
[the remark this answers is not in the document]
shotaronowhere: >  I think we'd need 4 seq inbox template combos (fee token x bufferable)

there's 2 sets of BridgeContracts (eth, erc20) and I added a delayBufferableSequencerInbox field to the BridgeContracts struct (none, delayBufferable).

that's what you mean right?

src/bridge/SequencerInbox.sol
gzeoneth, line 1: missing `addSequencerL2BatchDelayProof`, that is, calldata not from origin (DelayProof variant of `addSequencerL2Batch`)

shotaronowhere: Ok, we can add that method.

AFAIK, that method is only practically used for the rollup init message which is sequenced synchronously with deployment in the rollup creation process so no delay proof is required.

But for full flexibility with little down side, adding `addSequencerL2BatchDelayProof` probably makes sense

gzeoneth: actually need that for 3074 compatibility

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,177 @@
...
+        uint256 elapsed = end > start ? end - start : 0;
+        uint256 unexpectedDelay = prevDelay > threshold ? prevDelay - threshold : 0;
+        if (unexpectedDelay > elapsed) {
+            unexpectedDelay = elapsed;
+        }
+        // decrease the buffer saturating at zero
+        buffer = buffer > unexpectedDelay ? buffer - unexpectedDelay : 0;
+        // saturate at threshold
+        buffer = buffer > threshold ? buffer : threshold;
+        return buffer;
```
gzeoneth, line 48: nit: slightly cheaper
```suggestion
        if (buffer > unexpectedDelay) {
            buffer = buffer - unexpectedDelay;
            if (buffer > threshold) {
                return buffer;
            }
            return threshold;
        } else {
            return 0;
        }
```

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,177 @@
...
+        }
+        return buffer;
+    }
+
+    /// @notice Conditionally updates the buffer. Depletes if the delay is unexpected, otherwise replenishes if the buffer is depleted.
+    /// @param start The beginning reference point
+    /// @param end The ending reference point
+    /// @param buffer The buffer to be replenished
+    /// @param max The maximum buffer
+    /// @param replenishRateInBasis The amount to replenish the buffer per block in basis points.
```
gzeoneth, line 82: missing some param

gzeoneth (changes requested):
https://github.com/OffchainLabs/nitro-contracts/pull/163/files#r1565371423

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,177 @@
...
+        uint256 max,
+        uint256 replenishRateInBasis
+    ) internal pure returns (uint256) {
+        if (threshold < prevDelay) {
+            // deplete due to unexpected delay
+            buffer = deplete(start, end, buffer, prevDelay, threshold);
+        } else if (buffer < max) {
+            // replenish depleted buffer
+            buffer = replenish(start, end, buffer, max, replenishRateInBasis);
+        }
```
gzeoneth, line 98: while in delay mode (i.e. `threshold < prevDelay`), the sequencer would forgo all buffer replenishment
e.g. if the sequencer delayed `threshold + 1`, not only he would deplete 1 second of buffer, he also would not be able to get the `threshold * rate` worth of replenishment

I am ok with the implementation tho we can also always replenish so that it might be easier to specify the sequencer always have `rate` per second of delay buffer

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    /// @dev    Calculates the margin a sequenced message is below the delay threshold
+    ///         defining a `sync validity` window during which no delay proofs are required.
+    /// @notice Updates the time / block until no delay proofs are required.
+    /// @param blockNumber The block number when the synced message was created.
+    function updateSyncValidity(BufferData storage self, uint64 blockNumber) internal {
+        // saturating at uint64 max handles large threshold settings
+        self.syncExpiry = self.threshold > type(uint64).max - blockNumber
+            ? type(uint64).max
+            : blockNumber + self.threshold;
```
gzeoneth, line 154:
```suggestion
        uint256 expiry = blockNumber + self.threshold;
        self.syncExpiry = expiry > type(uint64).max
            ? type(uint64).max
            : uint64(expiry);
```

src/bridge/DelayBufferTypes.sol
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: BUSL-1.1
+
+import "./Messages.sol";
+
+pragma solidity >=0.6.9 <0.9.0;
+
+/// @notice Delay buffer and delay threshold settings
+/// @param threshold The maximum amount of blocks that a message is expected to be delayed
+/// @param max The maximum buffer in blocks
+/// @param period The period in blocks between replenishment
```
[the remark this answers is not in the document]
shotaronowhere: thanks, fixed

scripts/deploymentUtils.ts
```diff
@@ -115,6 +126,13 @@ export async function deployAllContracts(
     maxDataSize,
     reader4844,
     true,
+    false
+  ])
+  const erc20DelayBufferableSequencerInbox = await deployContract('SequencerInbox', signer, [
+    maxDataSize,
+    reader4844,
+    true,
+    false
```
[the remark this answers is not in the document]
shotaronowhere: thanks, fixed

src/bridge/DelayBuffer.sol
shotaronowhere: hmmm this suggested version would revert for `self.threshold > type(uint64).max - blockNumber`.

the original is written in a peculiar way, but it permits threshold = type(uint64), or otherwise very large value, to effectively disable the buffer setting a syncExpiry to type(uint64).max

src/bridge/DelayBuffer.sol
shotaronowhere: sounds good, included this logic

src/bridge/DelayBuffer.sol
shotaronowhere: always replenish is a simpler policy, I went ahead and simplified the buffer update logic to always replenish.

src/bridge/DelayBuffer.sol
gzeoneth: but threshold is uint64

gzeoneth: not sure why it would revert, threshold is uint64 and expiry (`blockNumber + self.threshold`) is uint256

shotaronowhere: I went ahead and made a further simplification which remove the need to update the syncexpiry all together. I think it's a good logical simplification

gzeoneth (commented):
LGTM, some nit about naming and documentation
Great work!

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,125 @@
...
+        uint256 buffer,
+        uint256 sequenced,
+        uint256 threshold,
+        uint256 max,
+        uint256 replenishRateInBasis
+    ) internal pure returns (uint256) {
+        uint256 elapsed = end > start ? end - start : 0;
+        uint256 delay = sequenced > start ? sequenced - start : 0;
+        // replenishment rounds down
+        buffer += (elapsed * replenishRateInBasis) / BASIS;
```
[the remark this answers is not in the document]
shotaronowhere: added a comment about overflow safety

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,125 @@
...
+        self.prevBlockNumber = blockNumber;
+        self.prevSequencedBlockNumber = uint64(block.number);
+    }
+
+    /// @dev    The delay buffer can change due to pending depletion / replenishment due to previous delays.
+    ///         This function applies pending buffer changes to calculate buffer updates.
+    /// @notice Calculates the buffer changes up to the requested block number
+    /// @param self The delay buffer data
+    /// @param blockNumber The block number to process the delay up to
+    function pendingBufferUpdate(BufferData storage self, uint64 blockNumber)
```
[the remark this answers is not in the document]
shotaronowhere: fixed

src/bridge/DelayBuffer.sol
```diff
@@ -0,0 +1,125 @@
...
+    ///         Eg. when the sequencer recovers from an outage, it is able to wait threshold > finality time before queueing delayed messages to avoid L1 reorgs.
+    /// @notice Conditionally updates the buffer. Replenishes the buffer and depletes if delay is unexpected.
+    /// @param start The beginning reference point
+    /// @param end The ending reference point
+    /// @param buffer The buffer to be updated
+    /// @param sequenced The reference point when messages were sequenced
+    /// @param threshold The threshold to saturate at
+    /// @param max The maximum buffer
+    /// @param replenishRateInBasis The amount to replenish the buffer per block in basis points.
+    function bufferUpdate(
```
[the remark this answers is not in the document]
shotaronowhere: renamed this one too

godzillaba (dismissed):

gzeoneth:
this will be merged as part of bold 
https://github.com/OffchainLabs/bold/pull/617

gzeoneth:
Shipped with #193

PR #163, opened 2024-04-01
review: delay buffer

gzeoneth:
wip review, please see inline comments

### DIFF
--- a/src/bridge/AbsBridge.sol
+++ b/src/bridge/AbsBridge.sol
@@ -127,6 +127,7 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
         if (afterDelayedMessagesRead > 0) {
             delayedAcc = delayedInboxAccs[afterDelayedMessagesRead - 1];
         }
+        // @review should use accumulateSequencerInbox here
         acc = keccak256(abi.encodePacked(beforeAcc, dataHash, delayedAcc));
         sequencerInboxAccs.push(acc);
     }
--- a/src/bridge/DelayBuffer.sol
+++ b/src/bridge/DelayBuffer.sol
@@ -22,13 +22,13 @@ library DelayBuffer {
     /// @dev    This function handles synchronizing the sequencer inbox with the delayed inbox.
     ///         It is called by the sequencer inbox when a delayed message is sequenced, proving
     ///         any delays and updating the delay buffers. This function is only called when the
-    ///         sequencer inbox has been unexpectedly delayed (rare case)
+    ///         sequencer inbox has been unexpectedly delayed (unhappy case)
     /// @notice Synchronizes the sequencer inbox with the delayed inbox.
     /// @param self The delay buffer data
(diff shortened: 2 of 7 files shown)


src/bridge/DelayBuffer.sol
```diff
@@ -192,7 +200,9 @@ library DelayBuffer {
         uint64 maxBuffer,
         uint64 amountPerPeriod,
         uint64 period,
-        uint64 roundOff
+        uint64 roundOff 
+        // @review would rather blend (amountPerPeriod, period) together to a fixed unit
+        // instead of needing to keep track of 3 terms (amountPerPeriod, period, roundOff)
```
gzeoneth, line 205: there would still be rounding error but should be small enough to be negligible
we can always round down if we care about sequencer abusing the rounding

src/bridge/DelayBuffer.sol
```diff
@@ -130,6 +130,7 @@ library DelayBuffer {
         uint64 blockNumber,
         uint64 timestamp
     ) internal {
+        // @review we can also prevent this by limiting the value you can set as threshold
```
shotaronowhere, line 133: the thought is thresholdBlocks = type(uint64).max and thresholdSeconds = type(uint64).max effectively disables the buffer, so governance could turn off the buffer feature without upgrading the contract.

_setBufferConfig calls updateSyncValidity so saturation at type(unit64).max is handled

src/bridge/DelayBuffer.sol
shotaronowhere: you mean something like setting the replenish rate in the form of 1/period ? Or even a simplification which works for both blocks and seconds like if period = 12, then there is no round off.

gzeoneth: yes, to have sth like
```
BASIS = 10000
amountPerSecondInBasis = 1000 # equals to 0.1s per 1 second
```
then when we calculate how much to replenish, we always around off. The rounding should be small enough that is negligible, and always work against the sequencer (i.e. sequencer cannot gain more delay buffer by abusing the rounding)

src/bridge/AbsBridge.sol
```diff
@@ -127,6 +127,7 @@ abstract contract AbsBridge is Initializable, DelegateCallAware, IBridge {
         if (afterDelayedMessagesRead > 0) {
             delayedAcc = delayedInboxAccs[afterDelayedMessagesRead - 1];
         }
+        // @review should use accumulateSequencerInbox here
```
gzeoneth, line 130: this was not fixed in https://github.com/OffchainLabs/nitro-contracts/pull/160 yet

shotaronowhere: I implemented your suggestion when both sync and delay proofs were implemented.

when removing the sync proof, the auxiliary helper functions in the Messages.sol library were no longer used and removed. Therefore there is no longer an accumulateSequencerInbox function and hence the bridge was left unchanged.

PR #167, opened 2024-04-15
fix: comment for excess fee refund address

Jason-W123:


src/node-interface/NodeInterface.sol
```diff
@@ -18,7 +18,7 @@ interface NodeInterface {
      * @param deposit amount to deposit to sender in L2
      * @param to destination L2 contract address
      * @param l2CallValue call value for retryable L2 message
-     * @param excessFeeRefundAddress gasLimit x maxFeePerGas - execution cost gets credited here on L2 balance
+     * @param excessFeeRefundAddress (gasLimit x maxFeePerGas - execution) + maxSubmission - (autoredeem ? 0 : submission cost) gets credited here on L2 balance
```
gzeoneth, line 21:
```suggestion
     * @param excessFeeRefundAddress (gasLimit x maxFeePerGas - execution cost) + (maxSubmission - (autoredeem ? 0 : submission cost)) gets credited here on L2 balance
```

gzeoneth (changes requested):
we should also sync these change with the inbox e.g.
https://github.com/OffchainLabs/nitro-contracts/blob/90037b996509312ef1addb3f9352457b8a99d6a6/src/bridge/IInbox.sol#L80

src/bridge/IERC20Inbox.sol
```diff
@@ -53,7 +53,7 @@ interface IERC20Inbox is IInboxBase {
      * @param to destination L2 contract address
      * @param l2CallValue call value for retryable L2 message
      * @param maxSubmissionCost Max gas deducted from user's L2 balance to cover base submission fee
-     * @param excessFeeRefundAddress gasLimit x maxFeePerGas - execution cost gets credited here on L2 balance
+     * @param excessFeeRefundAddress (gasLimit x maxFeePerGas - execution cost) + (maxSubmission - (autoredeem ? 0 : submission cost)) gets credited here on L2 balance
```
gzeoneth, line 56: I believe this is actually a little bit more nuanced for the unsafe entrypoint since you might have l1callvalue < l2callvalue + fees, and you can only get fee refund for at most `l1callvaue - l2callvalue`

i.e.
`min(l1callvalue - l2callvalue, gasLimit x maxFeePerGas + maxSubmission) - execution cost - (autoredeem ? 0 : submission cost)`

I am thinking if we want to be so elaborated here or we should just put a high level description here while have the detail example in docs.

Jason-W123: I agree to put `max(0,min(l1callvalue - l2callvalue, gasLimit x maxFeePerGas + maxSubmission)) - execution cost - (autoredeem ? 0 : submission cost)` here

gzeoneth: you missed the `l1callvalue - l2callvalue` part
see if we have time to discuss in standup

gzeoneth (dismissed):

gzeoneth (approved):

PR #168, opened 2024-04-15
fix: 3074 support

gzeoneth:
This PR

- [x] Add deprecation notice to function affected by EIP-3074
- [x] Deprecate `sendL2MessageFromOrigin`
  - This is a breaking change but should be fine because as of now only an EOA can call this method, so we are not going to run into issue like breaking an immutable contract. Alternatively, we can have this entrypoint call into the non-origin entrypoint but may require changes in the nitro node to locate the data properly.

SequencerInbox `fromOrigin` entrypoints are not removed yet, it will be fixed before EIP-3074 ship and when we have more clarity on its implementation. Since those entrypoints are only callable by the sequencer it pose less risk than the permissionless `sendL2MessageFromOrigin` and hence not removed in this PR for now.

### DIFF
--- a/src/bridge/AbsInbox.sol
+++ b/src/bridge/AbsInbox.sol
@@ -6,6 +6,7 @@ pragma solidity ^0.8.4;
 
 import {
     DataTooLarge,
+    Deprecated,
     GasLimitTooLarge,
     InsufficientValue,
     InsufficientSubmissionCost,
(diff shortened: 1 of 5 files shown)


src/bridge/AbsInbox.sol
```diff
@@ -131,19 +132,8 @@ abstract contract AbsInbox is DelegateCallAware, PausableUpgradeable, IInboxBase
...
-    {
-        if (_chainIdChanged()) revert L1Forked();
-        // solhint-disable-next-line avoid-tx-origin
-        if (msg.sender != tx.origin) revert NotOrigin();
-        if (messageData.length > maxDataSize) revert DataTooLarge(messageData.length, maxDataSize);
-        uint256 msgNum = _deliverToBridge(L2_MSG, msg.sender, keccak256(messageData), 0);
-        emit InboxMessageDeliveredFromOrigin(msgNum);
-        return msgNum;
+    function sendL2MessageFromOrigin(bytes calldata) external pure returns (uint256) {
+        revert Deprecated();
```
PlasmaPower, line 136: Can't we just call the non-from-origin variant here? Afaik the only difference is gas usage, I don't think there's an aliasing difference or anything

shotaronowhere (approved):
LGTM

src/bridge/AbsInbox.sol
gzeoneth: we can, but as described in the PR I don't think we need to and would prefer the simplicity of deprecating the function entirely

PR #172, opened 2024-04-24
refactor: move machine hash logic to osp

gzeoneth:
This PR move `getStartMachineHash` and `getEndMachineHash` from `ChallengeLib` to `OneStepProofEntry`, this change would make the `ChallengeManager` not compatible with previously deployed OSP.


gvladika (approved):
lgtm

PR #173, opened 2024-04-24
feat: conditional osp

gzeoneth:
Depends on https://github.com/OffchainLabs/nitro-contracts/pull/172

This PR conditionally use different OSP to workaround a breaking change in the post-Stylus wasm machine.

Was https://github.com/OffchainLabs/stylus-contracts/pull/40

Resolves BLK-129


gvladika (approved):
lgtm

PR #175, opened 2024-04-25
Add unit test for conditional OSP

gvladika:


gzeoneth (approved):
would add an assertNe to make sure the 2 osp return different hashes, otherwise lgtm

PR #181, opened 2024-05-06
Fix the nitro CI builds.

eljobe:
For some reason, the master branch of nitro won't build the contracts if there are any remappings specified for a yul build.

This change removes the remappings for the yul build.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -14,6 +14,8 @@ src = 'yul'
 out = 'out/yul'
 libs = ['node_modules', 'lib']
 cache_path  = 'forge-cache/yul'
+remappings = []
+auto_detect_remappings = false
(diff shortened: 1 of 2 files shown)


gzeoneth:
sorry I didn't see this and already created https://github.com/OffchainLabs/nitro-contracts/pull/182
also want to merge into develop instead, but let me know if main without this hotfix is a blocker

eljobe:
> sorry I didn't see this and already created #182 also want to merge into develop instead, but let me know if main without this hotfix is a blocker

I mean, we do need the fix in the `contracts` submodule in the nitro repository or our CI fails.
See: https://github.com/OffchainLabs/nitro/pull/2274

PR #182, opened 2024-05-06
fix: yul remapping issue

gzeoneth:
https://github.com/foundry-rs/foundry/issues/7869


foundry.toml
```diff
@@ -8,14 +8,17 @@ optimizer = true
 optimizer_runs = 100
 via_ir = false
 solc_version = '0.8.9'
+remappings = ['ds-test/=lib/forge-std/lib/ds-test/src/','forge-std/=lib/forge-std/src/','@openzeppelin/contracts/=node_modules/@openzeppelin/contracts/','@openzeppelin/contracts-upgradeable/=node_modules/@openzeppelin/contracts-upgradeable/']
```
eljobe, line 11: Optional: I'd format this as one-per-line.

eljobe (approved):

eljobe:
Oh, one other thing. I noticed when I was trouble-shooting this, that:

```
forge config
```

Output considerably more remappings than the four which were listed in remappings.txt.

So, is it possible that it was augmenting that list through some sort of auto-detection and that hardcoding these 4 in the configuration file directly might have some adverse consequences? (Totally just being paranoid.)

shotaronowhere (approved):

gzeoneth:
> Oh, one other thing. I noticed when I was trouble-shooting this, that:
> 
> ```
> forge config
> ```
> 
> Output considerably more remappings than the four which were listed in remappings.txt.
> 
> So, is it possible that it was augmenting that list through some sort of auto-detection and that hardcoding these 4 in the configuration file directly might have some adverse consequences? (Totally just being paranoid.)



> forge config

Thanks, I believe `auto_detect_remappings` has always been enabled and it will add entries to `remappings`, but those in the static list would always be included. This PR should keep the expected behavior. (For YUL, remapping are irrelevant)

PR #184, opened 2024-05-13
MultiCallTest.sol: a non-standard contract for stylus compatibility test

tsahee:
This contract is not standard or pretty. It's created and used to have the same api as a simple rust/stylus smart-contract. I have excluded it from formatting tests as much as I could.


gzeoneth (approved):
LGTM as a mock

PR #187, opened 2024-05-22
feat: anytrust fast confirmation

gzeoneth:
Similar to https://github.com/OffchainLabs/bold/pull/348 but have more validation (e.g. staker and zombie) and require sequential confirmation. Does NOT support create node and fast confirm in the same call.

Also fixed CI lacking a contract size check and overriding RollupUserLogic to use a lower optimizer-run to workaround the contract size issue

`anyTrustFastConfirmer` MUST also be set as validator to function, otherwise the call would revert when trying to remove zombies; `setAnyTrustFastConfirmer` and `setValidator` need to be called separately due to poor event in this version and code size limit. 

All the change in this PR are temporary and would be superseded by https://github.com/OffchainLabs/nitro-contracts/pull/193 when BOLD ship.

resolves BLK-161


yahgwai (dismissed):

src/rollup/RollupCore.sol
```diff
@@ -77,6 +77,7 @@ abstract contract RollupCore is IRollupCore, PausableUpgradeable {
     uint64 internal constant GENESIS_NODE = 0;
 
     bool public validatorWhitelistDisabled;
+    address public anyTrustFastConfirmer;
```
godzillaba, line 80: should we move this below `_hostChainIsArbitrum` to keep storage layout the same

gvladika (approved):
lgtm

PR #191, opened 2024-06-14
Upgradeable cache manager

yahgwai:
Add an init function to allow the cache manager to be deployed behind a proxy


gvladika (commented):
Logic lgtm, though there are some lint issues. For failed Slither check, you can run locally `slither . --skip-assembly --triage-mode` and select CacheManager warnings to add to ignore list (this was not done at the time CacheManager was first implemented because Slither CI was broken at that point).

godzillaba (approved):

gzeoneth (approved):

PR #192, opened 2024-06-17
Add script to deploy cache manager in testnode

gvladika:
Add function in utilities that deploys cache managers and enables it in precompile
Add script and yarn action to enable calling it from testnode


gzeoneth (approved):
LGTM

PR #194, opened 2024-06-21
Specify address when caching Stylus program

PlasmaPower:


src/precompiles/ArbWasmCache.sol
```diff
@@ -16,11 +16,14 @@ interface ArbWasmCache {
     /// @return managers the list of managers.
     function allCacheManagers() external view returns (address[] memory managers);
 
-    /// @notice Caches all programs with the given codehash.
+    /// @dev Deprecated, replaced with cacheProgram
+    function cacheCodehash(bytes32 codehash) external;
```
joshuacolvin0, line 20: What about completely removing this now?  Only on testnet, so not the end of the world if some things break until updated

PlasmaPower: We want to maintain the ability to sync past the testnet's ArbOS 30 section (for now at least)

joshuacolvin0 (approved):
LGTM

tsahee (dismissed):
LGTM

src/chain/CacheManager.sol
```diff
@@ -107,17 +107,18 @@ contract CacheManager {
     }
 
     /// Places a bid, reverting if payment is insufficient.
-    function placeBid(bytes32 codehash) external payable {
+    function placeBid(address program) external payable {
         if (isPaused) {
             revert BidsArePaused();
         }
+        bytes32 codehash = program.codehash;
```
gzeoneth, line 114: can we either also make codehash a parameter and compare it with `program.codehash` or at least check ` program.codehash != keccak256("")` to make this safer?

src/chain/CacheManager.sol
```diff
@@ -31,7 +31,7 @@ contract CacheManager {
     error BidsArePaused();
     error MakeSpaceTooLarge(uint64 size, uint64 limit);
 
-    event InsertBid(bytes32 indexed codehash, uint192 bid, uint64 size);
+    event InsertBid(bytes32 indexed codehash, address program, uint192 bid, uint64 size);
```
gzeoneth, line 34:
```suggestion
    event InsertBid(bytes32 indexed codehash, address indexed program, uint192 bid, uint64 size);
```
maybe also index program address

gzeoneth (commented):
minor comments, otherwise lgtm

src/chain/CacheManager.sol
PlasmaPower: The reason why I didn't do this is because you can cache somebody else's program by using another one with the same codehash. If people search by program they might miss important events.

src/chain/CacheManager.sol
PlasmaPower: The latter check is implicit because activation will fail if it's not a Stylus program, but I'd be happy to add a codehash parameter if you think we should have one. That sounds reasonable to me.

gzeoneth: > latter check is implicit
sgtm then!

src/chain/CacheManager.sol
gzeoneth: make sense

gzeoneth (dismissed):

gzeoneth (dismissed):

gzeoneth (approved):

PR #219, opened 2024-07-29
Wait for the deployment of upgrade executor

ImJeremyHe:


scripts/deploymentUtils.ts
```diff
@@ -207,6 +207,7 @@ export async function deployAllContracts(
   )
   const rollupUser = await deployContract('RollupUserLogic', signer, [], verify)
   const upgradeExecutor = await deployUpgradeExecutor(signer)
+  await upgradeExecutor.deployTransaction.wait()
```
Sneh1999, line 210: @**Jeremy He**  why are we waiting for only the `UpgradeExecutor` contract to deploy and not for any other contracts

Sneh1999: nvm It makes sense to me now - deployContract is calling `wait` for other contracts

gzeoneth (approved):
LGTM

PR #224, opened 2024-08-05
Ability to set alternative transferor

yahgwai:


src/express-lane-auction/ExpressLaneAuction.sol
```diff
@@ -446,7 +464,14 @@ contract ExpressLaneAuction is
         ELCRound storage resolvedRound = latestResolvedRounds.resolvedRound(round);
 
         address resolvedELC = resolvedRound.expressLaneController;
-        if (resolvedELC != msg.sender) {
+        address transferor = transferorOf[resolvedELC].addr;
+        // can only be the transferor if one has been set
+        // otherwise we default to the express lane controller to do the transfer
+        if (transferor != address(0)) {
+            if (transferor != msg.sender) {
+                revert NotTransferor(transferor, msg.sender);
```
gzeoneth, line 472: should be consistant with the NotExpressLaneController error that also emit round number

src/express-lane-auction/ExpressLaneAuction.sol
```diff
@@ -431,6 +433,22 @@ contract ExpressLaneAuction is
         );
     }
 
+    /// @inheritdoc IExpressLaneAuction
+    function setTransferor(Transferor calldata transferor) external {
+        // if a transferor has already been set, it may be fixed until a future round
+        Transferor storage currentTransferor = transferorOf[msg.sender];
+        if (
+            currentTransferor.addr != address(0) &&
+            currentTransferor.fixedUntilRound > roundTimingInfo.currentRound()
```
gzeoneth, line 442: maybe its better to fix to a timestamp (and rounding up 1 round) instead? otherwise changing round time might make the fix duration change

yahgwai: I think that's fine - round duration changes might also warrant a change of fix time. Also changing to timestamp might mean that you cant guarantee that the next round will have a fixed transferor

PR #230, opened 2024-08-09
small `AbsBridge::enqueueSequencerMessage` gas optimization

godzillaba:


gzeoneth (changes requested):
LGTM, but let's hold from merging for now until we have a new bridge version

gzeoneth (approved):
LGTM

PR #234, opened 2024-08-15
ci: upload 4bytes

gzeoneth:
- [x] Upload 4bytes using `cast` in the last step of contract test CI. 
- [x] Add a test:update shortcut to update storage and 4bytes files

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -129,7 +129,7 @@ jobs:
 
       - uses: OffchainLabs/actions/run-nitro-test-node@main
         with:
-          nitro-testnode-ref: deneb-integration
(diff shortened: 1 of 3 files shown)


.github/workflows/contract-tests.yml
```diff
@@ -125,7 +129,7 @@ jobs:
 
       - uses: OffchainLabs/actions/run-nitro-test-node@main
         with:
-          nitro-testnode-ref: deneb-integration
+          nitro-testnode-ref: gzeoneth-patch-1
```
yahgwai, line 132: why do we need this special testnode ref?

gzeoneth: testing stuff, almost there

yahgwai (approved):

PR #237, opened 2024-08-15
Added int offset

yahgwai:


src/express-lane-auction/RoundTimingInfo.sol
```diff
@@ -88,7 +100,7 @@ library RoundTimingInfoLib {
         pure
         returns (uint64, uint64)
     {
-        uint64 roundStart = info.offsetTimestamp + info.roundDurationSeconds * round;
+        uint64 roundStart = uint64(info.offsetTimestamp + int64(info.roundDurationSeconds * round));
```
gzeoneth, line 103: while all the cast here are unsafe, they are generally safe under the assumption that 1) timestamp is well within the range of uint32 and 2) offset is before timestamp

however, the particular cast is unsafe even with those assumption with e.g. negative offset + round 0

gzeoneth (approved):
LGTM, commented on an edge case of `roundTimestamps`

src/express-lane-auction/RoundTimingInfo.sol
gzeoneth: Anyway I suppose [this](https://github.com/OffchainLabs/nitro-contracts/blob/5f288d76b1fd7cd3dff6960208217d9a982565b7/src/express-lane-auction/IExpressLaneAuction.sol#L246-L249) already explained the limitation of the `roundTimestamps` method that make this specific edge case not a concern,  but maybe we can consider to store the round where the last round change happened and revert `roundTimestamps` if a round before that is requested.

yahgwai: I'll add more documentation about this limitation

PR #240, opened 2024-08-26
Cargo Stylus Factory

chrisco512:
Draft for a Stylus factory contract that may be used by the `cargo stylus` CLI tool to deploy, activate, and run an arbitrary constructor in a single transaction.

In Stylus/WASM context we are unable to run Rust or WASM code during the deployment step. There is also a need to activate WASM programs, which is currently abstracted from users via the `cargo stylus` tool. Users can currently define in their Rust code an "init" method but it must be called after activation and there is a need to gate the function manually to a specific deployer address as well as add arbitrary state variables to gate other methods from being invoked before the contract is initialized. 

This contract enables a better developer experience. The intention is to deterministically deploy this contract using create2 on Arbitrum chains so that Cargo Stylus can use it to deploy. The Stylus SDK will add a future `#[constructor]` proc macro that will gate a defined method to being invoked by this contract alone during the deployment transaction.


src/stylus/CargoStylusFactory.sol
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.13;
+
+interface IArbWasm {
+    function activateProgram(
+        address program
+    ) external payable returns (uint16 version, uint256 dataFee);
+}
+
+contract CargoStylusFactory {
```
gzeoneth, line 10: missing a payable receive function to receive refund from constructor (if any)

src/stylus/CargoStylusFactory.sol
```diff
@@ -0,0 +1,123 @@
...
+        // Initialize the contract
+        initializeContract(newContractAddress, _constructorCalldata, msg.value);
+
+        return newContractAddress;
+    }
+
+    // Computes amount to refund and sends it to msg.sender
+    function refundExcessValue(uint256 _activationFee, uint256 _constructorValue) internal {
+        // Calculate value to forward to constructor
+        uint256 excessValue = msg.value - _activationFee - _constructorValue;
```
gzeoneth, line 67: can simply refund `address(this).balance`

src/stylus/CargoStylusFactory.sol
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        emit ContractActivated(_contract);
+
+        return dataFee;
+    }
+
+    // Internal function to initialize a Stylus contract by
+    // invoking its constructor
+    function initializeContract(
```
gzeoneth, line 110: if we were to consolidate the 2 public function, we might as well inline these instead of making internal functions

gzeoneth (changes requested):

src/stylus/CargoStylusFactory.sol
```diff
@@ -0,0 +1,123 @@
...
+        // Deploy the contract
+        address newContractAddress = deployContract(_bytecode);
+
+        // Activate the contract
+        uint256 activationFee = activateContract(newContractAddress, activationValue);
+
+        // Initialize the contract
+        initializeContract(newContractAddress, _constructorCalldata, _constructorValue);
+
+        refundExcessValue(activationFee, _constructorValue);
```
gzeoneth, line 42: this ignore any refund from the constructor call (if any), I think we can wrap a balance check before and after the call to constructor to know the net amount spent in the constructor and do refund appropriately

there is also an additional edge case where the constructor cost is negative (if it somehow received more fund from the constructor)

gzeoneth:
while there were some comment regarding constructor refund, maybe we can assume it will never happen like other CREATE2 factories

src/stylus/CargoStylusFactory.sol
gzeoneth: specifically, I mean if `dataFee < _activationValue` in
```
(, uint256 dataFee) = IArbWasm(ARB_WASM).activateProgram{value: _activationValue}(
            _contract
        );
```
I am not sure if that refund involves evm call tho

src/stylus/CargoStylusFactory.sol
```diff
@@ -0,0 +1,123 @@
...
+    // Internal function to activate a Stylus contract
+    function activateContract(
+        address _contract,
+        uint256 _activationValue
+    ) internal returns (uint256) {
+        (, uint256 dataFee) = IArbWasm(ARB_WASM).activateProgram{value: _activationValue}(
+            _contract
+        );
+
+        emit ContractActivated(_contract);
```
gligneul, line 103: I'm getting this error if the program was previously activated:
```
    (code: 3, message: execution reverted: error ProgramUpToDate(), data: Some(String("0xcc944bf2")))
```

godzillaba:
replaced by #294

PR #246, opened 2024-09-06
Fix ArbWASM precompile minInitGas() method return types

ganeshvanahalli:
Corresponding nitro PR- https://github.com/OffchainLabs/nitro/pull/2640
Part of NIT-2771


joshuacolvin0 (approved):
LGTM

PR #250, opened 2024-09-17
Handle "no zero transfer" tokens

godzillaba:
resolves BLK-253

looked to find places where zero value custom gas token (or rollup stake token) transfers could occur. these are marked in 2 commits: https://github.com/OffchainLabs/nitro-contracts/pull/250/commits/b69072d0dbb8a52ffcffd06103768db6262a50ad and https://github.com/OffchainLabs/nitro-contracts/pull/250/commits/ed56f917de47f13ecf6baa3b97bd67929208635b

after reviewing the token bridge contracts too, it seems that zero transfers are only _maybe_ possible in the creator if a zero gas price bid is set. This is fine though

mentioning https://github.com/OffchainLabs/nitro-contracts/pull/230, because this PR may be enough of a reason to merge it


gzeoneth (changes requested):
LGTM, thinking about release strategy might have to wait for bold to release first

test/foundry/ERC20Outbox.t.sol
```diff
@@ -6,7 +6,7 @@ import "./ERC20Bridge.t.sol";
 import "../../src/bridge/ERC20Bridge.sol";
```
yahgwai, line 6: Should we also have an inbox/bridge test for enqueueing a delayed message

gzeoneth (approved):
LGTM

PR #252, opened 2024-09-19
Support for rollup mode when using fee token

gvladika:
This feature adds support to run fee token based Orbit chains in rollup mode. Previously only Anytrust mode was supported when fee token is used. This is because ArbO assumes that the asset used to pay for batch posting on parent chain is the same asset which is used as native asset on the child chain. If batch poster spends 0.01 ETH to post batch on the parent chain, it will be reimbursed 0.01 ETH on the child chain. However this approach is not valid when fee token is used. Reimbursing batch poster with 0.01 fee token after it has spent 0.01 ETH would be fine only if exchange rate between ETH and fee token was 1:1. In practice this will not be the case, so we’re introducing a way to get the actual exchange rate on-chain and use it to correctly reimburse the batch poster.

This PR enables SequencerInbox to fetch the feeToken:nativeToken exchange rate from the external source called `feeTokenPricer` which is set by the chain owner. Pricer needs to implement a single function `getExchangeRate() returns(uint256)`. In practice, pricer can use Chainlink, amm twap, constant price, or any other approach to report exchange rate.


src/bridge/ISequencerInbox.sol
```diff
@@ -220,10 +229,33 @@ interface ISequencerInbox is IDelayedMessageProvider {
...
+// CHRIS: TODO: analyse possible effects of a malicious fee token pricer
+// CHRIS: TODO: should the interface be a view function?
+interface IFeeTokenPricer {
+    /**
+     * @notice Get the number of child chain's fee tokens per 1 parent chain's native token. Exchange rate must be
+     *         denominated in 18 decimals.
+     * @dev    For example, parent chain's native token is ETH, fee token is DAI. If price of 1ETH = 2000DAI, then function should return 2000*1e18.
+     *         If fee token is USDC instead and price of 1ETH = 2000USDC, function should still return 2000*1e18, no matter that USDC uses 6 decimals.
+     */
+    function getExchangeRate() external returns (uint256);
```
gzeoneth, line 260: instead of a getter that return spot rate, I think it make more sense to have a mutable method `spendingInEth(uint256 wei, address spender)` that take the number of eth spent as input and return the number of token spent. This allow the pricer to do more complex internal accounting if needed.

gzeoneth (approved):
LGTM

gzeoneth:
merge after https://github.com/OffchainLabs/nitro-contracts/pull/322

godzillaba (approved):

godzillaba (approved):

gzeoneth (approved):

PR #255, opened 2024-09-27
[NIT-2798] Add contracts for HostIO gas tests

gligneul:
These contracts will be used by the nitro tests that check the equivalence between stylus and EVM.

Related Nitro PR: https://github.com/OffchainLabs/nitro/pull/2708


eljobe (dismissed):
LGTM

eljobe (approved):
LGTM

gzeoneth (approved):
LGTM
This PR does make the foundry config misalign with the hardhat config, I think it is ok since we are only using the hardhat build for production and it seems unlikely the different solc version and evm target would causes discrepancy in our tests. Might put up another PR in the near future to pin the pragma solidity of all contracts tho.

PR #258, opened 2024-10-01
chore: pin pragma

gzeoneth:
Pin pragma to the expected version due to https://github.com/OffchainLabs/nitro-contracts/pull/255 removed the default solc in foundry config.


gzeoneth:
This might make importing nitro-contract to develop on another solc version much harder, maybe we should keep the interfaces and abstract contracts floating.

PR #259, opened 2024-10-02
fix: eip 7702

gzeoneth:
In the inbox and sequencer inbox, there are some `FromOrigin` methods which expect the calldata exists in the top-level transaction. This is currently guaranteed by checking if `tx.origin==msg.sender` and thus it must be an EOA and must be a top-level call. [EIP-7702](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md) changes this by allowing an EOA to contain code.

This PR make those methods compatible to EIP-7702 by introducing an additional `msg.sender.code.length==0` check.


PlasmaPower:
Should we have a library function to do the origin and code check? That'd help deduplicate code and make it easier to update in the future. It could be called `isCallerCertainlyEoa()` or something

src/libraries/GasRefundEnabled.sol
```diff
@@ -24,8 +25,7 @@ abstract contract GasRefundEnabled {
             startGasLeft += calldataWords * 6 + (calldataWords ** 2) / 512;
             // if triggered in a contract call, the spender may be overrefunded by appending dummy data to the call
             // so we check if it is a top level call, which would mean the sender paid calldata as part of tx.input
-            // solhint-disable-next-line avoid-tx-origin
-            if (msg.sender != tx.origin) {
```
yahgwai, line 28: Should also update the natspec dev comment as it references msg.sender == tx.origin

yahgwai (approved):
LGTM! Only minor suggestions

PR #263, opened 2024-11-14
chore: bold deploy script improvements

gzeoneth:
Various changes to the bold deployment script, no contract change.

- Allow overriding `ROLLUP_ADDRESS` for testnode integration
- Wait for deployments to be included in a block
- Allow the upgrade execution script to execute the upgrade with a private key
- Add `ANVILFORK` flag for fork testing
- Parse log properly to extract post upgrade informations
- Add more logging for debug
- Fixes an issue where wrong `currentInboxCount` is used for validation

This PR also changed the imported nitro-contract version, but will be deprecated and removed in https://github.com/OffchainLabs/nitro-contracts/pull/266


scripts/boldUpgradeCommon.ts
```diff
@@ -94,9 +98,9 @@ export const validateConfig = async (
   l1Rpc: providers.Provider
 ) => {
   // check all the config.contracts exist
-  if ((await l1Rpc.getCode(config.contracts.excessStakeReceiver)).length <= 2) {
```
yahgwai, line 97: why is this commented out? can it just be deleted if we dont need it?

scripts/executeBoldUpgrade.ts
```diff
@@ -83,60 +86,111 @@ async function perform(
...
+    timelockSigner
   )
   const boldAction = BOLDUpgradeAction__factory.connect(
     deployedContracts.boldAction,
-    timelockImposter
+    timelockSigner
   )
 
-  // what validators did we have in the old rollup?
+  // // what validators did we have in the old rollup?
```
yahgwai, line 111: double comment

scripts/executeBoldUpgrade.ts
```diff
@@ -315,32 +369,44 @@ async function checkOldRollup(params: VerificationParams) {
   ) {
     throw new Error('Old rollup was not upgraded')
   }
+
+  // using the reader factory here for typing
+  const rollupReaderContract = RollupReader__factory.connect(config.contracts.rollup, l1Rpc)
```
yahgwai, line 374: why does this rollupreader still exist? i thought we were getting rid of it

scripts/boldUpgradeCommon.ts
gzeoneth: ya I think we can just delete

scripts/executeBoldUpgrade.ts
gzeoneth: it is in the #268 PR, I don't want to mix the contract change with this PR but that will be merged after this one

PR #264, opened 2024-11-18
docs: add note regarding reorg

gzeoneth:


gzeoneth:
will be fixed with https://github.com/OffchainLabs/nitro-contracts/pull/265

PR #265, opened 2024-11-18
use create2 in BridgeCreator

godzillaba:
the salt includes all the `createBridge` args in case it is called with a `rollup` that wasn't created by `RollupCreator`. If all args aren't included in this scenario, you could frontrun a bridge creation with different arguments, preventing the proper deployment through the bridge creator.


src/rollup/BridgeCreator.sol
```diff
@@ -84,8 +86,18 @@ contract BridgeCreator is Ownable {
...
         ISequencerInbox.MaxTimeVariation calldata maxTimeVariation
     ) external returns (BridgeContracts memory) {
+        // make sure the rollup exists
+        // otherwise the rollup creator could be DOSed by frontrunning
+        if (rollup.code.length == 0) {
+            revert RollupCodeEmpty();
+        }
+
+        // use create2 salt to ensure deterministic addresses
+        bytes32 create2Salt = keccak256(abi.encode(adminProxy, rollup, nativeToken, maxTimeVariation));
```
godzillaba, line 96: technically `adminProxy` isn't required in the salt since it's a constructor arg. can't hurt to include it though IMO

godzillaba:
@gzeoneth your changes lgtm

src/rollup/BridgeCreator.sol
gzeoneth: using the entire msg.data

src/rollup/BridgeCreator.sol
```diff
@@ -107,11 +124,20 @@ contract BridgeCreator is Ownable {
...
         BufferConfig calldata bufferConfig
     ) external returns (BridgeContracts memory) {
+        // make sure the rollup exists
+        // otherwise the rollup creator could be DOSed by frontrunning
+        if (rollup.code.length == 0) {
+            revert RollupCodeEmpty();
+        }
+
+        // use create2 salt to ensure deterministic addresses
+        bytes32 create2Salt = keccak256(msg.data);
```
gzeoneth, line 134: we are expecting the rollup address is already reorg safe and controlled by the sender, so we don't need to have msg.sender in the salt

gzeoneth: alternatively this is probably cleaner and also safe
```suggestion
        bytes32 create2Salt = keccak256(abi.encode(msg.data, msg.sender));
```

PR #270, opened 2024-11-26
refactor: move hardcoded value to config

gzeoneth:
Note that the 4bytes of `RollupCreator.createRollup` and `RollupAdminLogic.initialize` will be changed, won't be an issue if we have the change with the v3 release because they was already different from v2.


src/rollup/BOLDUpgradeAction.sol
```diff
@@ -465,14 +463,6 @@ contract BOLDUpgradeAction {
...
         );
 
-        // verify
-        (uint256 _delayBlocks, uint256 _futureBlocks, uint256 _delaySeconds, uint256 _futureSeconds)
-        = ISequencerInbox(SEQ_INBOX).maxTimeVariation();
-        require(_delayBlocks == delayBlocks, "DelayBuffer: delayBlocks");
-        require(_delaySeconds == delaySeconds, "DelayBuffer: delaySeconds");
-        require(_futureBlocks == futureBlocks, "DelayBuffer: futureBlocks");
-        require(_futureSeconds == futureSeconds, "DelayBuffer: futureSeconds");
-
```
gzeoneth, line 475: have to remove these checks for contract code size, I don't think there are too much risk but arguable

scripts/config.ts.example
```diff
@@ -17,8 +17,12 @@ export const config = {
     owner: '0x1234123412341234123412341234123412341234',
     loserStakeEscrow: ethers.constants.AddressZero,
     chainId: ethers.BigNumber.from('13331370'),
+    minimumAssertionPeriod: 75,
+    validatorAfkBlocks: 201600,
     chainConfig:
       '{"chainId":13331370,"homesteadBlock":0,"daoForkBlock":null,"daoForkSupport":true,"eip150Block":0,"eip150Hash":"0x0000000000000000000000000000000000000000000000000000000000000000","eip155Block":0,"eip158Block":0,"byzantiumBlock":0,"constantinopleBlock":0,"petersburgBlock":0,"istanbulBlock":0,"muirGlacierBlock":0,"berlinBlock":0,"londonBlock":0,"clique":{"period":0,"epoch":0},"arbitrum":{"EnableArbOS":true,"AllowDebugPrecompiles":false,"DataAvailabilityCommittee":false,"InitialArbOSVersion":10,"InitialChainOwner":"0x1234123412341234123412341234123412341234","GenesisBlockNum":0}}',
+    minimumAssertionPeriod: ethers.BigNumber.from('75'),
+    validatorAfkBlocks: ethers.BigNumber.from('201600'),
```
godzillaba, line 25:
```suggestion
```

these are redundant

scripts/rollupCreation.ts
```diff
@@ -303,6 +303,8 @@ async function _getDevRollupConfig(
       loserStakeEscrow: ethers.constants.AddressZero,
       chainId: JSON.parse(chainConfig)['chainId'],
       chainConfig: chainConfig,
+      minimumAssertionPeriod: 75,
```
godzillaba, line 306: should this be `0` since it's dev?

godzillaba (approved):
lgtm - just one (or two) small change

scripts/rollupCreation.ts
gzeoneth: I will leave it to have the same behavior as previously for now, not sure changing it will break testnode or sth

src/rollup/BOLDUpgradeAction.sol
```diff
@@ -433,17 +442,6 @@ contract BOLDUpgradeAction {
             PROXY_ADMIN_SEQUENCER_INBOX.upgrade(sequencerInbox, IMPL_SEQUENCER_INBOX);
         }
 
-        // verify
```
yahgwai, line 436: Is there manual checklist somewhere that we could add these to?

yahgwai (approved):

src/rollup/BOLDUpgradeAction.sol
gzeoneth: not yet but we can make something

PR #272, opened 2024-12-04
parent and child insurance stubs

godzillaba:


src/insurance/Parent.sol
srya-b, line 1: is the assumption that the `signedCommitment` in `buy` also encodes the values being committed to alongside the commitment + signature?

godzillaba: yeah, just pushed up a commit that includes those. the entire `struct SequencerCommitment` is signed over

srya-b (approved):

PR #275, opened 2024-12-10
chore: effectively disableing delay buffer

gzeoneth:
If `isDelayBufferable` is set to false, it is hard to re-enable without doing contract upgrade due to the flag is stored in an immutable for gas optimization. We temporarily disable delay buffer by using a large threshold value so that `isSynced` return true and will not require a delay proof when posting batches.
```
    function isSynced(
        BufferData storage self
    ) internal view returns (bool) {
        return block.number - self.prevBlockNumber <= self.threshold;
    }
```
the previous config for sepolia is not large enough because `prevBlockNumber` might be 0 after an upgrade

### DIFF
--- a/scripts/files/configs/local.ts
+++ b/scripts/files/configs/local.ts
@@ -43,8 +43,8 @@ export const local: Config = {
     maxDataSize: 117964,
     isDelayBufferable: true,
     bufferConfig: {
-      max: 14400,
-      threshold: 300,
+      max: 2 ** 32, // effectively disableing and will be enabled later
+      threshold: 2 ** 32, // 365 days, effectively disableing and will be enabled later
       replenishRateInBasis: 500,
     },
   },
--- a/scripts/files/configs/local.ts
+++ b/scripts/files/configs/local.ts
@@ -44,7 +44,7 @@ export const local: Config = {
     isDelayBufferable: true,
     bufferConfig: {
       max: 2 ** 32, // effectively disableing and will be enabled later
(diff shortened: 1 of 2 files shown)


scripts/files/configs/local.ts
```diff
@@ -43,8 +43,8 @@ export const local: Config = {
     maxDataSize: 117964,
     isDelayBufferable: true,
     bufferConfig: {
-      max: 14400,
-      threshold: 300,
+      max: 2 ** 32, // effectively disableing and will be enabled later
+      threshold: 2 ** 32, // 365 days, effectively disableing and will be enabled later
```
godzillaba, line 47:
```suggestion
      threshold: 2 ** 32, // effectively disableing and will be enabled later
```

scripts/files/configs/sepolia.ts
```diff
@@ -36,8 +36,8 @@ export const sepolia: Config = {
     maxDataSize: 117964,
     isDelayBufferable: true,
     bufferConfig: {
-      max: hoursToBlocks(24 * 365), // 365 days, effectively disableing and will be enabled later
-      threshold: hoursToBlocks(24 * 365), // 365 days, effectively disableing and will be enabled later
+      max: 2 ** 32, // effectively disableing and will be enabled later
+      threshold: 2 ** 32, // 365 days, effectively disableing and will be enabled later
```
godzillaba, line 40:
```suggestion
      threshold: 2 ** 32, // effectively disableing and will be enabled later
```

godzillaba (approved):

PR #284, opened 2024-12-17
ArbOS11To32UpgradeTest

diegoximenes:
Resolves NIT-2975

Adds a test contract that has a function that uses mcopy, an instruction that is supported in ArbOS 32 but not in ArbOS 11.
It also updates audit-ci.jsonc, with changes that are already in the develop branch.

To be used in https://github.com/OffchainLabs/nitro/pull/2837

Attention to the base branch.
Current version of develop is not compatible with nitro, so a `pre-bold` branch, checked out from main, was created to support changes that should be introduced to nitro while bold is not fully integrated to it.

### DIFF
--- a/audit-ci.jsonc
+++ b/audit-ci.jsonc
@@ -65,6 +65,16 @@
     // Server-Side Request Forgery in axios
     "GHSA-8hc4-vh64-cxmj",
     // Regular Expression Denial of Service (ReDoS) in micromatch
-    "GHSA-952p-6rrq-rcjv"
+    "GHSA-952p-6rrq-rcjv",
+    // cookie accepts cookie name, path, and domain with out of bounds characters
+    "GHSA-pxg6-pf52-xh8x",
+    // Elliptic's verify function omits uniqueness validation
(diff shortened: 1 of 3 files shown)


gzeoneth (approved):
LGTM, I think the plan is not to merge into main but tag v2.1.2 on this branch (only if you need a npm release, if not I might even not bother to tag it but nitro can simply pin to this commit)

Can you also create the same PR on `develop` so we have this mock also available in v3 (bold)?

diegoximenes:
> LGTM, I think the plan is not to merge into main but tag v2.1.2 on this branch (only if you need a npm release, if not I might even not bother to tag it but nitro can simply pin to this commit)

I don't need a npm release so I will just merge it.

> Can you also create the same PR on `develop` so we have this mock also available in v3 (bold)?

I created the PR with develop as the base branch: https://github.com/OffchainLabs/nitro-contracts/pull/286

pre-bold doesn't have merge protection if the PR was not approved BTW.
I don't have access to nitro-contract repo settings, could you set this protection?

gligneul (approved):

PR #294, opened 2025-01-29
Stylus deployer

yahgwai:


src/stylus/StylusDeployer.sol
```diff
@@ -0,0 +1,177 @@
...
+            // anywhere the address of this contract is seen the same init data and value were used.
+            salt = initSalt(salt, initData, initValue);
+        }
+
+        address newContractAddress = deployContract(bytecode, salt);
+        bool shouldActivate = requiresActivation(newContractAddress);
+        if (shouldActivate) {
+            // ensure there will be enough left over for init
+            uint256 activationValue = msg.value - initValue;
+            ARB_WASM.activateProgram{value: activationValue}(newContractAddress);
```
godzillaba, line 72:
```suggestion
            // ensure there will be enough left over for init
            // activateProgram will return unused value back to this contract without an EVM call
            uint256 activationValue = msg.value - initValue;
            ARB_WASM.activateProgram{value: activationValue}(newContractAddress);
```

godzillaba: or something like this to explain what's happening with the value

src/stylus/StylusDeployer.sol
```diff
@@ -0,0 +1,177 @@
...
+            uint256 bal = address(this).balance;
+            if (bal != 0) {
+                // the caller must be payable, even if they dont expect to receive any balance since it value can be forced into
+                // this contract via selfdestruct
+                (bool sent,) = payable(msg.sender).call{value: bal}("");
+                if (!sent) {
+                    revert RefundExcessValueError(bal);
+                }
+            }
+        }
```
godzillaba, line 100:
```suggestion
            uint256 bal = address(this).balance;
            if (bal != 0) {
                // the caller must be payable, even if they dont expect to receive any balance since it value can be forced into
                // this contract via selfdestruct
                (bool sent,) = payable(msg.sender).call{value: bal}("");
                if (!sent) {
                    revert RefundExcessValueError(bal);
                }
            }
```

godzillaba: i don't think `if (shouldActivate || msg.value != initValue)` is necessary

godzillaba (commented):
haven't dug into the tests yet but the deployer looks pretty good

src/stylus/StylusDeployer.sol
```diff
@@ -0,0 +1,177 @@
...
+    function deploy(
+        bytes calldata bytecode,
+        bytes calldata initData,
+        uint256 initValue,
+        bytes32 salt
+    ) public payable returns (address) {
+        if (salt != 0) {
+            // if a salt was supplied, hash the salt with init value and init data. This guarantees that
+            // anywhere the address of this contract is seen the same init data and value were used.
+            salt = initSalt(salt, initData, initValue);
```
gzeoneth, line 64: I think we should not include `initValue` for consistency with evm CREATE2 factories. Stylus contract init is like constructor initialisation_code for normal evm contract, and CREATE2 does not salt with the deployment value. i.e. 
```
address = keccak256(0xff + sender_address + salt + keccak256(initialisation_code))[12:]
```

src/stylus/StylusDeployer.sol
```diff
@@ -0,0 +1,177 @@
...
+        } else if (initValue != 0) {
+            // if initValue exists init data should too
+            revert InitValueButNotInitData();
+        }
+
+        // refund any remaining value if:
+        // - activation can return some to this contract
+        // - some activation value was supplied but not used
+        if (shouldActivate || msg.value != initValue) {
+            uint256 bal = address(this).balance;
```
gzeoneth, line 91: This is not safe against reentrancy, the inner deployment will get all the refund. We can instead check pre and post balance.

gzeoneth (changes requested):

src/stylus/StylusDeployer.sol
yahgwai: you thinking that the balance check is cheap, and that if there's balance the caller can just have it?

src/stylus/StylusDeployer.sol
```diff
@@ -0,0 +1,177 @@
...
+        address newContractAddress;
+        if (salt != 0) {
+            assembly {
+                newContractAddress := create2(0, add(bytecode, 0x20), mload(bytecode), salt)
+                // bubble up the revert if there was one
+                if and(iszero(newContractAddress), not(iszero(returndatasize()))) {
+                    let p := mload(0x40)
+                    returndatacopy(p, 0, returndatasize())
+                    revert(p, returndatasize())
+                }
```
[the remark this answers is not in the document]
yahgwai: my assembly foo isnt good enough, you got a suggestion?

src/stylus/StylusDeployer.sol
yahgwai: Updated

src/stylus/StylusDeployer.sol
yahgwai: Now using:
```
uint256 bal = msg.value - dataFee - initValue;
```

src/stylus/StylusDeployer.sol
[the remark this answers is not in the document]
godzillaba: lol no the assembly is good i was just thinking like this:

```solidity
address newContractAddress;
if (salt != 0) {
    assembly {
        newContractAddress := create2(0, add(bytecode, 0x20), mload(bytecode), salt)
    }
} else {
    assembly {
        newContractAddress := create(0, add(bytecode, 0x20), mload(bytecode))
    }
}

assembly {
    // bubble up the revert if there was one
    if and(iszero(newContractAddress), not(iszero(returndatasize()))) {
        let p := mload(0x40)
        returndatacopy(p, 0, returndatasize())
        revert(p, returndatasize())
    }
}
```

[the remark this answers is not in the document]
yahgwai: ah, I didnt know I would have access to the return data in a second assembly block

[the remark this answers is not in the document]
yahgwai: Updated

godzillaba (approved):
LGTM

chrisco512 (approved):

godzillaba:
should close #240

gzeoneth (approved):
LGTM

gzeoneth (approved):

PR #295, opened 2025-01-31
Add a BigMap contract for testing the storage trie

eljobe:
This contracts makes it easy to add a bunch of storage slots in a single transation, and then in a separate transaction to free a bunch of those slots and add some others.

The idea is that this sort of operation will pass with the deterministic flag on the storage db implementation set and fail when it is not set.

Part of NIT-3068

### DIFF
--- a/src/mocks/BigMap.sol
+++ b/src/mocks/BigMap.sol
@@ -1,0 +1,23 @@
+// Copyright 2025, Offchain Labs, Inc.
+// For license information, see:
+// https://github.com/OffchainLabs/nitro/blob/master/LICENSE.md
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.0;
+
+contract BigMap {
+    mapping(uint256 => uint256) public data;
+    uint256 size;
+
+    function clearAndAddValues(uint256 clear, uint256 add) external {
+        uint256 i = size;
+        while (i < size + add) {
+            data[i] = 8675309;
+            i++;
+        }
+        size = i;
+        for (uint256 j = 0; j < clear; j++) {
+            data[j] = 0;
+        }
+    }
+}


eljobe:
@PlasmaPower, I'm not sure what all the CI failures mean. Did I do something wrong?

PlasmaPower:
Some of the failures are because the contract isn't formatted correctly. I'd fix that and see what's still failing. The audit is probably just not passing on the base branch, so I think we can merge it without that one passing.

src/mocks/BigMap.sol
```diff
@@ -0,0 +1,32 @@
...
+// https://github.com/OffchainLabs/nitro/blob/master/LICENSE.md
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.0;
+
+contract BigMap {
+    mapping(uint256 => uint256) public data;
+		uint256 size;
+
+    function storeValues(uint256 count) external {
```
PlasmaPower, line 12: Not that it really matters, but could we just use clearAndAddValues with clear=0 instead of having this function?

eljobe: Sure. Will do.

How does one format a Solidity contract?

PlasmaPower: In the contracts repo you run `yarn prettier:solidity`

eljobe: Fixed the formatting and made it just a single method.

PlasmaPower (approved):
LGTM

gzeoneth:
sorry, I missed this PR when releasing 2.1.3; do we need a release tag for this mock or we are ok with just merging this into the permanent `v2-main` branch?

gzeoneth (approved):

PlasmaPower:
We don't need a release for this, this is just for nitro unit tests so it just needs to be in the nitro pin

PR #297, opened 2025-02-03
Update BOLD prepare script for orbit

godzillaba:
* deploys ERC20 versions of contracts if appropriate
* only deploys and sets 4844 reader if existing sequencer inbox uses one
* new `custom.ts` config file
* small updates to bold execution script

related: https://github.com/OffchainLabs/orbit-actions/pull/39


godzillaba:
#298 fixes CI on `develop`

gzeoneth (approved):
LGTM

PR #308, opened 2025-03-03
fix: yarn audit ci

gzeoneth:

PR #311, opened 2025-03-14
Add methods to enable/disable and check the EIP-7623 feature

eljobe:
This precompile will allows the feature to be off by default, but toggled on and
off via a single call to the ArbOwner contract.

Part of: NIT-3139


eljobe (commented):
LGTM

gzeoneth (approved):
LGTM, bypassing CI because simple interface change. Will fix CI in #308

PR #314, opened 2025-03-20
Add methods to enable/disable and check the EIP-7623 feature (#311)

eljobe:
* Add methods to enable/disable and check the EIP-7623 feature

This precompile will allows the feature to be off by default, but toggled on and off via a single call to the ArbOwner contract.

Part of: NIT-3139

* format: yarn format

* docs: note applicable arbos version

* chore: move getter to ArbOwnerPublic

* chore: mark getter as view

This PR just brings these methods into `v2-main`. Already merged into `develop`


gzeoneth (changes requested):
checking with nitro team to see if we can get rid of using `v2-main` and use `main` with some legacy abi added instead

eljobe:
@gzeoneth and @godzillaba, the nitro team (@tsahee) is working to figure out how to reduce our dependencies on the `v2-main` branch in https://github.com/OffchainLabs/nitro/pull/3048 of the `nitro-contracts` repository so that we can truly consider it to be "deprecated" and only use it in tests of our legacy staker (until such a time that the staker can also be removed, because we're 100% confident in BoLD.)

But, for now, we need to get a release cut for Pectra support, and the `nitro` repository is still depending on `nitro-contracts/v2-main`. Would it be possible to prioritize the review and merging of this PR so we can unblock progress on Pectra even before we get @tsahee's change in?

gzeoneth (approved):

PR #315, opened 2025-03-20
Add a method for testing redeeming multiple retryable tickets at once

eljobe:
This was previously added in the private repo, and I have. similar PR open for putting the change in the `develop` branch.

Related to: NIT-3010


gzeoneth (changes requested):
checking with nitro team to see if we can get rid of using `v2-main` and use `main` with some legacy abi added instead

joshuacolvin0:
> checking with nitro team to see if we can get rid of using `v2-main` and use `main` with some legacy abi added instead

Yes, that is absolutely what we want, but really need this PR merged ASAP because it is currently blocking Nitro development.

eljobe:
> Ok, please run `yarn format` to fix CI before merge

On this branch `yarn format` makes massive changes across the set of solidity contracts. 77 files changed. I'm hesitant to commit that to the branch. But, if I should, please confirm.

PR #316, opened 2025-03-24
Adds the setWasmMaxSize precompile function to ArbOwner

eljobe:
Chain owners that wish to allow larger wasm contracts can increase the size via an ArbOwner call on their chain.

Part of: NIT-3173


src/precompiles/ArbOwner.sol
```diff
@@ -161,6 +161,11 @@ interface ArbOwner {
         uint16 limit
     ) external;
 
+    /// @notice Sets the maximum size of the uncompressed wasm code in bytes
+    function setWasmMaxSize(
```
gzeoneth, line 165: do we have a getter? do we need one?

eljobe: Similar to the other "setWasm*" setters in this contract, there isn't a need for getters. The effects of setting this are read internally from the underlying storage by the node software.

Now, there could be a usability argument here. "How are chain owners supposed to know the current value when they want to set the new value?" Typically, these values show up in log messages when they are exceeded. So, the chain owners usually know if they need to update them from their current values based on these logs. But, maybe we should open a feature request to make this more transparent either through contract calls, or exposing RPC endpoints for inspecting the current values for this storage. That's really for product to decide, and I don't think it should block this review.

gzeoneth (approved):
LGTM, overriding unrelated failing CIs

PR #319, opened 2025-03-27
SelfDestruct helper contracts for unit testing SelfDestruct

amsanghi:


gzeoneth (approved):
https://github.com/OffchainLabs/nitro-contracts/pull/320 keeping these change insync between v2-main and develop

PR #332, opened 2025-04-28
Add MintNativeToken and BurnNativeToken calls

eljobe:
These will allow arbOwner authorized addresses to mint some amount of the chain's native gas token to or burn some amount of the chain's native gas token from a given address.

Chain owners *should* deploy a non-upgradable intermediary contract, the address of which is added to the set of arbOwner addresses and then add any addresses which should be allowed to mint and burn the native gas token to call corresponding methods on the proxy contract. The implementation of those intermediary contract methods should not allow to and from address arguments, and instead use the address of the caller as the target for minting and burning.

Resolves: NIT-3297

### DIFF
--- a/src/precompiles/ArbOwner.sol
+++ b/src/precompiles/ArbOwner.sol
@@ -211,13 +211,13 @@ interface ArbOwner {
      * @notice Mints some amount of the native gas token for this chain to the
      * given address.
      */
-    function mintNativeToken(address to, uint64 amount) external;
+    function mintNativeToken(address to, uint256 amount) external;
 
     /**
      * @notice Burns some amount of the native gas token for this chain from the
      * given address.
      */
-    function burnNativeToken(address from, uint64 amount) external;
+    function burnNativeToken(address from, uint256 amount) external;
 
     /**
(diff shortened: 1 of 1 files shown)


src/precompiles/ArbOwner.sol
```diff
@@ -207,6 +207,18 @@ interface ArbOwner {
         string calldata chainConfig
     ) external;
 
+    /**
+     * @notice Mints some amount of the native gas token for this chain to the
+     * given address.
+     */
+    function mintNativeToken(address to, uint64 amount) external;
```
diegoximenes, line 214: Why amount is not an uint256?

eljobe: Good catch. It should be uint256.

diegoximenes (approved):

godzillaba:
i think it's worth considering whether we want the burner to be able to burn from any account or only their own account.

currently the only way a contract's balance can be reduced is if it makes external calls with value. having the burner be able to burn from arbitrary accounts would change this. this might be weird for app devs

PR #333, opened 2025-04-28
Bold upgrade action QoL improvements

TucksonDev:
This PR contains QoL improvements in the BoLD upgrade action process:
- instead of deploying all logic contracts, it relies on an already deployed RollupCreator (with v3.1.0 template contracts) to grab the logic contracts
- it adds more validation of the config used in custom.ts
- it fixes a few verification issues that didn't involve potential extra Orbit configurations (e.g. anytrustFastConfirmer)


scripts/deployment.ts
```diff
@@ -1,21 +1,34 @@
...
 import { deployAllContracts, _isRunningOnArbitrum } from './deploymentUtils'
-import { maxDataSize } from './config'
+import { maxDataSize as defaultMaxDataSize } from './config'
 
 import { ArbSys__factory } from '../build/types'
 
 async function main() {
-  const [signer] = await ethers.getSigners()
+  let signer;
+  if (process.env.DEPLOYER_PRIVKEY !== undefined) {
```
TucksonDev, line 10: Adding here 2 optional env variables so we can deploy a RollupCreator from the testnode by specifying a `DEPLOYER_PRIVKEY` and a `MAX_DATA_SIZE`. They both fall back to the previous values used.

scripts/files/templatesV3.1.ts
TucksonDev, line 1: All template addresses hardcoded. Decided to hardcode the single contracts instead of the BridgeCreator in case we change the templates directly in the BridgeCreator in the future.

At the end of the file, we have the templates that are deployed in the testnode as part of the new boldupgrade process.

scripts/prepareBoldUpgrade.ts
```diff
@@ -32,31 +33,26 @@ async function main() {
     configNetworkName + 'DeployedContracts.json'
   )
 
-  // if the deployed contracts exists then we load it and combine
```
TucksonDev, line 35: I removed this block that was supposed to handle partial deployments, since we are now deploying only one contract (and reusing deployed templates)

scripts/files/templatesV3.1.ts
```diff
@@ -0,0 +1,243 @@
...
+  challengeManagerTemplate: string
+  osp: string
+  rollupCreator: string
+}
+
+export const templates: {
+  [key: number]: CreatorTemplates
+} = {
+  // Ethereum
+  1: {
```
gzeoneth, line 29: can we have some check in the upgrade action to validate these are correct on the rpc provided?

scripts/boldUpgradeCommon.ts
```diff
@@ -142,18 +146,27 @@ export const validateConfig = async (
   }
 
   // check all the settings exist
+  if (config.settings.challengeGracePeriodBlocks === 0) {
+    throw new Error('challengeGracePeriodBlocks is 0')
+  }
```
gzeoneth, line 151: I think this can be 0

scripts/boldUpgradeCommon.ts
```diff
@@ -142,18 +146,27 @@ export const validateConfig = async (
...
   }
   if (config.settings.chainId === 0) {
     throw new Error('chainId is 0')
   }
+  if (config.settings.minimumAssertionPeriod === 0) {
+    throw new Error('minimumAssertionPeriod is 0')
+  }
+  if (config.settings.validatorAfkBlocks === 0) {
+    throw new Error('validatorAfkBlocks is 0')
+  }
```
gzeoneth, line 169: This can be 0, 0 is to disable the validator afk feature.

gzeoneth (commented):
Looking good, small comments

scripts/files/templatesV3.1.ts
TucksonDev: Added in https://github.com/OffchainLabs/nitro-contracts/pull/333/commits/c7f6b048e36fbef5ec83d8e03ad012f745802e3a

scripts/boldUpgradeCommon.ts
TucksonDev: Oh right, thanks. Changed.

gzeoneth (approved):
LGTM

gzeoneth:
This is also tagged as v3.1.0-qol

PR #334, opened 2025-04-29
Mint-Burn gas token privileged contract

TucksonDev:
This PR adds a contract that will serve as a proxy for certain privileged accounts to call the new `mintNativeToken` and `burnNativeToken` functions of the ArbOwner precompile.

This privileged accounts will be managed by the chain owner, who must be the deployer of this contract.

This contract must be set as a chain owner of the chain to be able to call ArbOwner functions

Closes BLK-331

### DIFF
--- a/src/bridge/IMinterBurnerForwarder.sol
+++ b/src/bridge/IMinterBurnerForwarder.sol
@@ -1,0 +1,27 @@
+// Copyright 2021-2022, Offchain Labs, Inc.
+// For license information, see https://github.com/OffchainLabs/nitro-contracts/blob/main/LICENSE
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.0;
+
+interface IMinterBurnerForwarder {
+    /**
+     * @notice Mints some amount of the native gas token for this chain to the calling address.
+     * @dev This function calls mintNativeToken in the ArbOwner precompile, so this contract must also be a chain owner.
+     *      No events are emitted in this function, since the ArbOwner precompile already emits OwnerActs()
+     * @param amount The amount of native gas token to mint
+     */
+    function mintNativeToken(
+        uint256 amount
+    ) external;
+
+    /**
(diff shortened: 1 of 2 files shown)


src/bridge/IMinterBurnerForwarder.sol
```diff
@@ -0,0 +1,27 @@
+// Copyright 2021-2022, Offchain Labs, Inc.
+// For license information, see https://github.com/OffchainLabs/nitro-contracts/blob/main/LICENSE
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity ^0.8.0;
+
+interface IMinterBurnerForwarder {
```
godzillaba, line 7: perhaps we could add `BURNER_ROLE()` and `MINTER_ROLE()` to the interface

src/bridge/MinterBurnerForwarder.sol
```diff
@@ -0,0 +1,55 @@
...
+            _grantRole(BURNER_ROLE, burners[i]);
+        }
+    }
+
+    /// @inheritdoc IMinterBurnerForwarder
+    function mintNativeToken(
+        uint256 amount
+    ) external onlyRole(MINTER_ROLE) {
+        ARB_OWNER.mintNativeToken(msg.sender, amount);
+    }
```
godzillaba, line 47: we should do this to burn too
```suggestion
    function mintNativeToken(
        address to,
        uint256 amount
    ) external onlyRole(MINTER_ROLE) {
        ARB_OWNER.mintNativeToken(to, amount);
    }
```

src/bridge/IMinterBurnerForwarder.sol
TucksonDev: Oops, missed that. Thanks for pointing it out. Added.

src/bridge/MinterBurnerForwarder.sol
TucksonDev: Done

PR #339, opened 2025-05-06
feat: ERC20MigrationOutbox

gzeoneth:
# ERC20MigrationOutbox

## Overview
This PR introduces the `ERC20MigrationOutbox` contract, a new component designed to facilitate the migration of native tokens from the Arbitrum bridge to an external bridge on custom fee token orbit chains.

## Purpose
Some custom fee token orbit chains may want to have their native token managed by an external bridge rather than the built-in native bridge. This contract enables permissionless migration of native bridge collateral without requiring changes to the vanilla outbox implementation, while also preserving the message passing and deposit functionality of the native bridge.

## Technical Requirements
- Must be allowed as an outbox in conjunction with the vanilla outbox contract
- Nonzero value withdrawals via the native bridge (ArbSys) must be disabled on the child chain to prevent funds and messages from being stuck as the native bridge will not hold collaterals

Resolves BLK-336


godzillaba (approved):

TucksonDev (approved):
lgtm, added a few nits

PR #343, opened 2025-05-12
Update ExpressLaneAuction currentRound method comments

Jason-W123:


gzeoneth (approved):

src/express-lane-auction/IExpressLaneAuction.sol
```diff
@@ -226,6 +226,10 @@ interface IExpressLaneAuction is IAccessControlEnumerableUpgradeable, IERC165Upg
 
     /// @notice The current auction round that we're in
     ///         Bidding for control of the next round occurs in the current round
+    /// @dev On inactive chains without new block production, this method may return 
+    ///      stale round information using the last known block timestamp. For chains 
+    ///      with infrequent block creation, consider using off-chain timestamp calculations 
+    ///      or implementing an alternative round tracking mechanism to ensure accuracy.
```
gzeoneth, line 232:
```suggestion
    /// @dev On inactive chains without new block production, this method may return
    ///      stale round information using the last known block timestamp. For chains
    ///      with infrequent block creation, consider using off-chain timestamp calculations
    ///      or implementing an alternative round tracking mechanism to ensure accuracy.
```

PR #344, opened 2025-05-12
Constraint based pricing gas dimension trace tests

relyt29:
This PR adds tests to the contracts repo that are used in abigen/solgen in nitro to do system_tests of multidimensional gas tracking. This work is related to https://github.com/OffchainLabs/nitro/pull/3229 and https://github.com/OffchainLabs/go-ethereum/pull/457.

None of this code is related to the other contracts in this repo, and all of the solidity is only test solidity, not intended to be used outside of integration testing. The code is isolated into the gas-dimensions folder, and I added a foundry build target separate from the rest of the code (which is also necessary because this code will not compile with the yul optimizer set to on).

The main nitro repo can then call the new yarn build command in the package.json to compile the test contracts, and will add the compiled test contracts into the solgen folder in the nitro repo.

the scripts should also still be runnable using forge script so long as you pass the profile correctly e.g. something like 

```bash
./target/bin/nitro --dev;
FOUNDRY_PROFILE=gas-dimensions forge script gas-dimensions/script/Sstore.s.sol -vvvvv --private-key "nitro --dev environment private key here" --slow --broadcast --rpc-url http://127.0.0.1:8547 --priority-gas-price "1000000000" --with-gas-price "2000000000" -g "10000"  --chain-id "412346"
```


relyt29:
Closing in favor of #345

PR #349, opened 2025-05-20
feat: ArbNativeTokenManager events

gzeoneth:


src/precompiles/ArbNativeTokenManager.sol
```diff
@@ -11,6 +11,16 @@ pragma solidity >=0.4.21 <0.9.0;
  *         Available in ArbOS version 41
  */
 interface ArbNativeTokenManager {
+    /**
+     * @notice Emitted when some amount of the native gas token is minted by a NativeTokenOwner.
+     */
+    event NativeTokenMinted(address indexed minter, address indexed to, uint256 amount);
```
eljobe, line 17: Won't minter and to always be the same address?

src/precompiles/ArbNativeTokenManager.sol
```diff
@@ -11,6 +11,16 @@ pragma solidity >=0.4.21 <0.9.0;
...
+
+    /**
+     * @notice Emitted when some amount of the native gas token is burned by a NativeTokenOwner.
+     */
+    event NativeTokenBurned(address indexed burner, address indexed to, uint256 amount);
```
eljobe, line 22: Shouldn't the name of the target of the event be`from` instead of `to`?

eljobe (changes requested):

src/precompiles/ArbNativeTokenManager.sol
gzeoneth: oh right will fix

PR #351, opened 2025-05-22
Native token management rename

tsahee:
fixes: NIT-3389

This names describes the true function - enabling management of native token (add minters) and not enabling the native token itself

pulled by: https://github.com/OffchainLabs/nitro/pull/3264


eljobe (changes requested):
I believe this PR should have been opened against the `develop` branch and not the `main` branch. See the configuration for the submodule-pin-check: https://github.com/OffchainLabs/nitro/blob/44154b58ac4cd1456857de761fe56e0e006868c0/.github/workflows/submodule-pin-check.yml#L33

eljobe (approved):

godzillaba (approved):

PR #353, opened 2025-06-03
Custom da legacy

Tristan-Wilson:


src/osp/OneStepProverHostIo.sol
```diff
@@ -236,6 +307,52 @@ contract OneStepProverHostIo is IOneStepProver {
 
                 extracted = kzgProof[64:96];
             }
+        } else if (inst.argumentData == 3) {
```
tsahee, line 310: same comments as validate

PR #357, opened 2025-06-18
Add CustomDA proof validation interface

Tristan-Wilson:
- Introduce ICustomDAProofValidator interface for extensible DA proof validation
- Implement ReferenceDAProofValidator with hash verification and chunk extraction
- Integrate CustomDA validation into OneStepProverHostIo for preimage type 3
- Add comprehensive test coverage for reference validator implementation


src/osp/OneStepProverHostIo.sol
```diff
@@ -29,6 +30,15 @@ contract OneStepProverHostIo is IOneStepProver {
...
 
+    ICustomDAProofValidator public customDAValidator;
+
+    function setCustomDAValidator(
+        ICustomDAProofValidator _validator
+    ) external {
+        // TODO: Add appropriate access control
+        customDAValidator = _validator;
+    }
+
```
gzeoneth, line 41: we don't keep any storage in the OSP contract, so instead I think we have 2 options
1. store the var in the rollup and have the OSP read from the rollup
2. make this immutable and set in the constructor, but we have to consider how it works with the rollup creator

gzeoneth: I would prefer to make this immutable because this allow more deterministic behavior, otherwise the rollup owner may change the DA contract and make a assertion invalid.

gzeoneth:
will work on the review comments

yahgwai (approved):

yahgwai (dismissed):

gzeoneth (approved):
LGTM

PR #358, opened 2025-06-24
use create2 for factory deployment

godzillaba:
Implement CREATE2 deployment, also making BridgeCreator and ValidatorWalletCreator immutable.

### DIFF
--- a/scripts/compareBytecodes.ts
+++ b/scripts/compareBytecodes.ts
@@ -9,7 +9,6 @@ interface ComparisonResult {
   match: boolean
   hardhatLength: number
   foundryLength: number
-  difference: string
 }
 
 function getHardhatArtifacts(): Map<string, string> {
@@ -29,6 +28,9 @@ function getHardhatArtifacts(): Map<string, string> {
         const content = JSON.parse(fs.readFileSync(fullPath, 'utf8'))
         if (content.bytecode && content.bytecode !== '0x') {
           const contractName = file.replace('.json', '')
+          if (artifacts.has(contractName)) {
+            throw new Error(`Duplicate contract name found: ${contractName}`)
+          }
           artifacts.set(contractName, content.bytecode)
         }
       }
@@ -46,7 +48,7 @@ function getFoundryArtifacts(): Map<string, string> {
   const artifacts = new Map<string, string>()
   const outDir = path.join(__dirname, '..', 'out')
 
-  function scanDirectory(dir: string, basePath: string = '') {
+  function scanDirectory(dir: string) {
     const files = fs.readdirSync(dir)
 
     for (const file of files) {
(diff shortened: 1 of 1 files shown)


scripts/deploymentUtils.ts
```diff
@@ -102,15 +122,38 @@ export async function deployContract(
...
+  fac: ContractFactory,
+  deploymentArgs: Array<any>,
+  salt = ethers.constants.HashZero,
+  overrides?: Overrides
+): Promise<Contract> {
+  if (hexDataLength(salt) !== 32) {
+    throw new Error('Salt must be a 32-byte hex string')
+  }
+
+  const FACTORY = '0x4e59b44847b379578588920cA78FbF26c0B4956C'
```
gzeoneth, line 135: let's read this from env but default to 0x4e
we can also auto deploy when the factory does not exists yet, maybe gate that with a flag

src/rollup/BridgeCreator.sol
```diff
@@ -43,11 +43,13 @@ contract BridgeCreator is Ownable {
     }
 
     constructor(
+        address initialOwner,
```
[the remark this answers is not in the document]
godzillaba: f4ba76f0f934b41d5d120c4c3336ca1187bdb741

scripts/deploymentUtils.ts
godzillaba: i added a link to the factory's github repo instead. IMO auto deploy is probably not worth it since it's an uncommon case. happy to go ahead with it if you disagree

godzillaba:
Fixing CI

scripts/compareBytecodes.ts
```diff
@@ -0,0 +1,228 @@
...
+      const fullPath = path.join(dir, file)
+      const stat = fs.statSync(fullPath)
+
+      if (stat.isDirectory()) {
+        scanDirectory(fullPath)
+      } else if (file.endsWith('.json')) {
+        const content = JSON.parse(fs.readFileSync(fullPath, 'utf8'))
+        if (content.bytecode && content.bytecode !== '0x') {
+          const contractName = file.replace('.json', '')
+          artifacts.set(contractName, content.bytecode)
```
godzillaba, line 32: if we have two contracts with the same name but different paths, we could get a false positive / negative without this change

```suggestion
          if (artifacts.has(contractName)) {
            throw new Error(`Duplicate contract name found: ${contractName}`)
          }
          artifacts.set(contractName, content.bytecode)
```

scripts/compareBytecodes.ts
```diff
@@ -0,0 +1,228 @@
...
+        const content = JSON.parse(fs.readFileSync(fullPath, 'utf8'))
+        if (
+          content.bytecode &&
+          content.bytecode.object &&
+          content.bytecode.object !== '0x'
+        ) {
+          const contractName = file.replace('.json', '')
+          // Include the path in the key for Foundry artifacts
+          const key = basePath ? `${basePath}/${contractName}` : contractName
+          artifacts.set(key, content.bytecode.object)
```
godzillaba, line 72: why include the basePath? why not only use contractName?

gzeoneth: will get rid of it since we are already assuming contract names are unique

PR #363, opened 2025-07-01
chore: disable metahash and align hardhat foundry

godzillaba:
sepolia deployment still being verified


gzeoneth:
pushed a few changes to make it cleaner and align the hardhat vs foundry build a bit better; we are still unable to replicate those contracts that require overridden optimizer runs setting.

PR #373, opened 2025-07-29
chore: use arbos 41 precompiles pin

gzeoneth:


yahgwai (dismissed):

TucksonDev (dismissed):
LGTM.
I was a bit confused about the commit used, but it's because it's not in the v41 branch: https://github.com/OffchainLabs/nitro-precompile-interfaces/commits/b0a772ab165e1879f76ad915490230ec32b44b64/

PR #376, opened 2025-08-27
Updated precompiles version

alexandrosfilios:


eljobe:
We shouldn't need to do this. The blockchain engineering team can take care of updating the nitro-contracts repository.

PR #420, opened 2026-03-19
feat: add rpc polling interval to deployment script

alxdca:
This PR adds a POLLING_INTERVAL env var that can be used to configure ethers polling interval (defaults to 4s) in the deployment script.

### DIFF
--- a/scripts/deployment.ts
+++ b/scripts/deployment.ts
@@ -17,6 +17,10 @@ async function main() {
     signer = signers[0]
   }
 
+  if (process.env.POLLING_INTERVAL !== undefined) {
+    signer.provider.pollingInterval = Number(process.env.POLLING_INTERVAL)
+  }
+
   const maxDataSize =
     process.env.MAX_DATA_SIZE !== undefined
       ? Number(process.env.MAX_DATA_SIZE)
--- a/scripts/deploymentUtils.ts
+++ b/scripts/deploymentUtils.ts
@@ -185,7 +185,7 @@ export async function create2(
     data: concat([salt, data]),
     ...overrides,
   })
-  await tx.wait(2)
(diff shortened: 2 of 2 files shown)


scripts/deploymentUtils.ts
```diff
@@ -185,7 +185,7 @@ export async function create2(
     data: concat([salt, data]),
     ...overrides,
   })
-  await tx.wait(2)
+  await tx.wait(Number(process.env.CREATE2_CONFIRMATIONS ?? 2))
```
yahgwai, line 188: Are you having trouble with the current code? I'd be interested to know the situations where 2 confs is failing

scripts/deployment.ts
```diff
@@ -17,6 +17,10 @@ async function main() {
     signer = signers[0]
   }
 
+  if (process.env.POLLING_INTERVAL !== undefined) {
+    signer.provider.pollingInterval = Number(process.env.POLLING_INTERVAL)
```
yahgwai, line 21: Out of interest, why do you need this? Do you set it really low and try to speed up deployments?

scripts/deploymentUtils.ts
alxdca: It's useful in test scenarios with anvil where interval mining is disabled, only auto-mining, i.e. mining a block when a tx arrives, is active. With 2 confs required, we need another transaction and this blocks the test indefinitely.

scripts/deployment.ts
alxdca: Default polling time is 4s. Since we deploy up to 26 contracts, lowering it to 100ms can save quite some time on fast parent chains.

yahgwai (dismissed):

PR #427, opened 2026-04-22
[DO NOT MERGE] [MEL] Changes to Rollup contracts

TucksonDev:
!! WIP - DO NOT MERGE !!

> [!NOTE]
> This is a preliminary version, and can completely change, based on changes in the design or the implementation of the feature in nitro. 

This PR contains changes to support MEL in the Rollup contracts. Changes to the OSP contracts are added in #432 

## Key changes

- The target of the next assertion to create is not an Inbox position anymore. We now use the hash of the block previous to where the current assertion is created
- Onchain inbox accumulators are not used anymore, so now assertions are hashed only using the previous assertion hash and the current assertion data
- GlobalState contains now 4 new fields:
  - MELState hash
  - Next message hash
  - Msg count
  - Executed Msg count
- Setting a new MELConfig in the Rollup config allow it to set a new Inbox and SequencerInbox

## Pending work

- [x] Overflowing assertions
- [x] `nextParentChainBlockHash` to use on initialization
- [X] OSP contracts (handled in #432 )
- [ ] Fix e2e tests (once a nitro version is ready)
- [ ] Rollup update process

## Nitro reference PRs

- https://github.com/OffchainLabs/nitro/pull/4643


src/rollup/IRollupAdmin.sol
```diff
@@ -65,6 +65,11 @@ interface IRollupAdmin {
     /// @dev Challenge manager was set
     event ChallengeManagerSet(address challengeManager);
 
+    /// @dev MELConfig was set
+    event MELConfigSet(
+        uint64 indexed melVersion, address indexed inbox, address indexed sequencerInbox, uint64 activationBlock
```
ganeshvanahalli: shouldn't `melVersion` be uint16? as defined in MELState.sol?

ganeshvanahalli (changes requested):

src/rollup/BOLDUpgradeAction.sol
```diff
@@ -387,7 +387,6 @@ contract BOLDUpgradeAction {
             layerZeroBigStepEdgeHeight: BIGSTEP_LEAF_SIZE,
             layerZeroSmallStepEdgeHeight: SMALLSTEP_LEAF_SIZE,
             genesisAssertionState: genesisAssertionState,
-            genesisInboxCount: inboxMaxCount,
```
TucksonDev, line 390: Since the BOLDUpgradeAction depends on the current contracts, we need to remove this field. I'm wondering if we should remove this action from this branch now, or move it to a different repo.

PR #428, opened 2026-04-22
[ResourceConstraintManager] Set expiryTimestamp on first use

TucksonDev:
This PR modifies the ResourceConstraintManager so the `expiryTimestamp` is set to 2 years from the moment one of the `set` functions is called.


TucksonDev:
The "CI/Contract tests" check is failing due to formatting. I reverted the format changes to not include anything else than what's strictly needed (format changes are unrelated to code in this PR)

PR #439, opened 2026-06-30
Add TipCollectionToggler contract

TucksonDev:
This PR adds a TipCollectionToggler contract that allows a trusted party to enable/disable tip collection in a chain (available from ArbOS 60), during a specified time. This contract should be added as a chain owner.

It also updates the pin of the `precompiles` submodule to v60.0.0, so it includes the ArbOwner method that needs to be called.


src/chain/TipCollectionToggler.sol
```diff
@@ -0,0 +1,59 @@
...
+        expiryTimestamp = block.timestamp + ACTIVATION_DURATION;
+        emit Activated(expiryTimestamp);
+    }
+
+    /// @notice Removes the contract from the list of chain owners after the expiry timestamp
+    function revoke() external onlyActivated {
+        if (block.timestamp < expiryTimestamp) {
+            revert NotExpired();
+        }
+        ARB_OWNER.removeChainOwner(address(this));
```
godzillaba, line 51: nit but maybe we should add events to `revoke` and `setCollectTips`?

godzillaba: i know ArbOwner emits them too but can't hurt

godzillaba (changes requested):
LGTM! just a nit about events

we should probably wait for a precompiles release before merging to develop though

src/chain/TipCollectionToggler.sol
TucksonDev: Thanks! The removeChainOwner method already emits a specific `ChainOwnerRemoved` in nitro (since ArbOS 60). But I've added a CollectTipsUpdated(bool) event for the `setCollectTips` method, since nitro only emits the generic `OwnerActs` event. 
What do you think?

godzillaba: personally i'd just add a redundant event but this sgtm too. the `ChainOwnerRemoved` event alone doesn't tell you _how_ the chain owner was removed which i guess could be a reason to include it

TucksonDev: That's fair yeah. I added a simple `Revoked()` event

godzillaba (approved):
LGTM assuming CI passes

I pushed one commit to hide a slither issue

PR #443, opened 2026-07-28
chore: fix audit:ci and harden dependency installs

yahgwai:
Makes `audit:ci` pass.

- **Refreshed `yarn.lock`** within existing version ranges.
- **Added two `resolutions`**: `hardhat-deploy/axios` to `>=0.33.0 <1` (scoped, so `eth-gas-reporter` keeps its own axios; floating so future 0.x backports land on a refresh), and `undici` to `^6.27.0` (global, as `hardhat` and `hardhat-verify` share one hoisted copy).
- **Rewrote the `audit-ci.jsonc` allowlist**: removed entries that no longer match anything, and added new ones.
- **Moved CI to node 24**, `setup-node` to v4, audit matrix to `[24]`: the package lock refresh pulls `@nomicfoundation/edr` (`engines >= 20`, via `hardhat` 2.29.0) and `brace-expansion` 5.0.8 (`engines "20 || >=22"`, via `@types/minimatch` → `minimatch@*`), and yarn treats an `engines` mismatch as a hard error, so every job died at `yarn install`. 18 and 20 are both EOL, 24 is the active LTS, and `setup-node@v3` was already being force-migrated to the node 24 runtime.

Closes #441, #442, #433.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,5 +108,9 @@
     "typechain": "^8.3.2",
     "typescript": "^4.9.5",
     "yarn-audit-fix": "^10.1.1"
+  },
+  "resolutions": {
+    "hardhat-deploy/axios": ">=0.33.0 <1",
+    "undici": "^6.27.0"
   }
 }
--- a/.github/workflows/audit-ci.yml
+++ b/.github/workflows/audit-ci.yml
@@ -26,7 +26,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install node_modules
-        uses: OffchainLabs/actions/node-modules/install@main
+        uses: OffchainLabs/actions/node-modules/install@ed09b0d87a23153e0f95cf69495e0dd8c1997cef  # main as of 2026-07-28
         with:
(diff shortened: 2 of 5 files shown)


eljobe (approved):

package.json
```diff
@@ -108,5 +108,9 @@
     "typechain": "^8.3.2",
     "typescript": "^4.9.5",
     "yarn-audit-fix": "^10.1.1"
+  },
+  "resolutions": {
+    "hardhat-deploy/axios": ">=0.33.0 <1",
+    "undici": "^6.27.0"
```
TucksonDev, line 114: undici comes from hardhat-verify, which uses ^v5.14, and here we force ^v6. It should be fine, since it looks like it's only used for stable request/fetch APIs, but something to keep in mind.

TucksonDev (approved):
LGTM, one minor flag (probably no need to change)

package.json
yahgwai: yeh, ci is passing so it should be fine

